PR #6, opened 2020-05-12
Sambrett/unittesting

sam-brett:
I believe this completes tickets #127 and #128


src/test/java/uk/co/ramp/record/SeirRecordTest.java
```diff
@@ -0,0 +1,94 @@
...
+        Assert.assertEquals(time, seirRecord.getTime());
+        Assert.assertEquals(s, seirRecord.getS());
+        Assert.assertEquals(e, seirRecord.getE());
+        Assert.assertEquals(i, seirRecord.getI());
+        Assert.assertEquals(r, seirRecord.getR());
+
+    }
+
+    @Test
+    public void testTestToString() {
```
johnnonweiler: Why does this (and other similar methods) have the prefix 'testTest'?

src/test/java/uk/co/ramp/record/SeirRecordTest.java
```diff
@@ -0,0 +1,94 @@
...
+        String expected = String.format("SeirRecord{time=%d, s=%d, e=%d, i=%d, r=%d}", time, s, e, i, r);
+        Assert.assertEquals(expected, testRecord.toString());
+    }
+
+    @Test
+    public void testTestEquals() {
+
```
johnnonweiler, line 60: I would expect this test to include comparison with an unequal record.

src/test/java/uk/co/ramp/record/SeirRecordTest.java
```diff
@@ -0,0 +1,94 @@
...
+        SeirRecord seirRecord = useSecondContructor();
+        int hash1 = testRecord.hashCode();
+        int hash2 = seirRecord.hashCode();
+
+        // identical data = identical hash
+        Assert.assertEquals(hash1, hash2);
+
+        seirRecord = new SeirRecord(time, s, e, i, 0);
+
+        hash1 = testRecord.hashCode();
```
johnnonweiler: This line (setting hash1 again) is redundant.

src/main/java/uk/co/ramp/Properties.java
```diff
@@ -0,0 +1,45 @@
...
+    private static StandardProperties standardProperties;
+    private static DiseaseProperties diseaseProperties;
+    private static PopulationProperties populationProperties;
+
+    private Properties() {
+
+    }
+
+    public static StandardProperties getStandardProperties() throws IOException {
+        if (standardProperties == null) {
```
johnnonweiler, line 24: Why are you using lazy loading for these properties?
(I would expect you'd try loading them all at the beginning, and fail early if any are missing.)

src/main/java/uk/co/ramp/people/Person.java
```diff
@@ -112,33 +119,34 @@ public void randomExposure(int t) {
...
             LOGGER.info("Person with id: {} has been randomly exposed at time {}", getId(), t);
             updateStatus(EXPOSED, t);
+        } else {
+            String message = String.format("The person with id: %d should not be able to transition from %s to %s", id, status, EXPOSED);
+            LOGGER.error(message);
+            throw new InvalidStatusTransitionException(message);
         }
     }
 
     int getDistributionValue(double mean, ProgressionDistribution p) {
```
johnnonweiler, line 129: I don't understand the `getDistributionValue` method.  The value of `sample` appears to be unused?  (And then '14' seems to be a magic number.)

src/test/java/uk/co/ramp/people/PersonTest.java
```diff
@@ -0,0 +1,266 @@
...
+        } catch (InvalidStatusTransitionException e) {
+            Assert.assertThat(appender.getOutput(), containsString(String.format("Changing status from RECOVERED for person.id %d is not a valid transition", id)));
+            throw e;
+        }
+    }
+
+    @Test
+    public void notNow() {
+        int time = rnd.nextInt(0, 100);
+        int flatTransition = (int) diseaseProperties.getMeanTimeToInfected();
```
johnnonweiler, line 153: It would be more obvious that flatTransition was not equal to 10, if it was set (or checked) explicitly here.

src/test/java/uk/co/ramp/people/PersonTest.java
```diff
@@ -0,0 +1,266 @@
...
+    public void randomExposure() {
+        int time = rnd.nextInt(0, 1000);
+        person.randomExposure(time);
+        Assert.assertThat(appender.getOutput(), containsString(String.format("Person with id: %d has been randomly exposed at time %d", id, time)));
+    }
+
+
+    @Test(expected = InvalidStatusTransitionException.class)
+    public void invalidExposure() {
+        int time = rnd.nextInt(0, 1000);
```
johnnonweiler, line 172: This highlights an interesting design choice:  A person who has already been exposed cannot be exposed again.  (I would have guessed that anyone could be exposed, it just wouldn't effect some people.)

src/test/java/uk/co/ramp/people/PersonTest.java
```diff
@@ -0,0 +1,266 @@
...
+            String message = String.format("The person with id: %d should not be able to transition from %s to %s", person.getId(), person.getStatus(), EXPOSED);
+            Assert.assertThat(appender.getOutput(), containsString(message));
+            throw e;
+        }
+
+    }
+
+    @Test
+    public void getGaussianDistributionValue() {
+        int n = 10000;
```
johnnonweiler, line 193: I don't understand this test (but I haven't spent long trying to figure it out).

johnnonweiler: (and similarly for the following couple of tests)

src/test/java/uk/co/ramp/people/PopulationGeneratorTest.java
```diff
@@ -0,0 +1,264 @@
...
+        Assert.assertEquals(e, result.get(EXPOSED).intValue());
+        Assert.assertEquals(i, result.get(INFECTED).intValue());
+        Assert.assertEquals(r, result.get(RECOVERED).intValue());
+
+    }
+
+    //        PopulationGenerator.createCumulative();
+    @Test
+    public void testCreateCumulative() {
+
```
johnnonweiler, line 72: This test is hard to follow, and I looked at the createCumulative method, and it also looks a bit strange.  I think the createCumulative method and its test would be good things to talk through sometime.

src/test/java/uk/co/ramp/people/PopulationGeneratorTest.java
```diff
@@ -0,0 +1,264 @@
...
+    }
+
+
+    @Test
+    public void testFindAgeSimple() {
+
+        Random rnd = new Random();
+
+        Map<Integer, Double> b = new HashMap<>();
```
johnnonweiler, line 120: These are not the most helpful variable names :-)

johnnonweiler (approved):
Looks like you've been busy - making a bunch of other little improvements at the same time as adding lots of tests!

Reviewing this on a commit-by-commit basis didn't work very well - the commit message explained one change, but then there were other apparently unrelated changes in the commits, as well as one cryptic commit message, "Adding test for".  So, I reviewed all the commits together instead, and I've made a few minor comments.

Generally, it all looks good to me.

src/main/java/uk/co/ramp/Properties.java
vinopm: +1

src/test/java/uk/co/ramp/record/SeirRecordTest.java
sam-brett: I think I must have missed that out of the commit. Will fix.

src/main/java/uk/co/ramp/Properties.java
sam-brett: This class has been removed.

src/main/java/uk/co/ramp/people/Person.java
sam-brett: sample is a random variable taken from the chosen distribution (Gaussian, Exp etc) which is then rounded to an int, but with a maximum duration of 14 days. This is a magic number but I'll add a todo because its something I need to discuss with Sibylle and Louise.

src/test/java/uk/co/ramp/people/PersonTest.java
sam-brett: Added assertion

src/test/java/uk/co/ramp/people/PersonTest.java
sam-brett: I may be wrong, but from the SEIR papers I've read, the 'exposed' state is more like a pre-symptomatic, infected but not infectious state. Happy to be corrected.

src/test/java/uk/co/ramp/people/PersonTest.java
sam-brett: They are just checking the rng is getting something around the mean value for a large number of runs. Just to make sure the distributions behave.

src/test/java/uk/co/ramp/record/SeirRecordTest.java
sam-brett: Strangely, on refreshing this page they have appeared below

src/test/java/uk/co/ramp/record/SeirRecordTest.java
sam-brett: clearly I'm testing twice as hard.... *headslap*... will fix

src/test/java/uk/co/ramp/people/PopulationGeneratorTest.java
sam-brett: Yeah, i think the blood/caffeine balance had shifted too far in favour of blood. I'll add comments. 
Effectively, the population data I got was not in a cumulative form, but I didn't want to change the raw data so I am creating a cumulative set on the fly. Some of this stuff can probably be removed in the future anyway. The final value is forced to 1 to avoid 0.9999 coming up on occasion.

vinopm (approved):

vinopm:
@sam-brett I'm ready with my changes, will commit and create a PR once you have merged this one to develop as i'm working on the same branch

sam-brett:
Thanks @vinopm. I'm starting to refactor the code around the new compartments, but if you see anything that you think could be improved, please feel free to suggest. This has been thrown together so more eyes/brains are always good. Especially as I'm feeling a bit frazzled this week

vinopm:
@sam-brett nice! I have several things that I'd like to suggest to improve readability and promote cleanliness. I'll suggest these once we've got a stable baseline build in develop to work with, rather than packing more changes into this PR. Good job on the PR btw.

sam-brett:
@vinopm great. Maybe we can have a more "in the weeds" dev chat tomorrow after the catch up. I've begun a bit of a refactor to start moving logic out of the ContactRunner and Person class because they're growing a little too fat for my liking.

vinopm:
@sam-brett sure, sounds good. Completely agree with you that we should refactor those classes into smaller separate components.

PR #7, opened 2020-05-14
Unit tests for contact, io, utils packages + some refactoring

vinopm:
1. Addition of unit tests for contact, io, utilities packages
2. Some refactoring from introducing Immutables library (https://github.com/immutables/immutables). Immutables provide a nice way to create immutable value objects via fluent builders. I've implemented this to reduce boilerplate code and to increase readibility.
3. Unit tests use fluent assertions through use of AssertJ (https://github.com/joel-costigliola/assertj-core). I quite like using fluent assertions instead of the standard junit assertions. Let me know what you think. 
4. Changes to parse JSON and CSV directly to immutable value objects instead of for example parsing csv contents row by row ourselves. I've tried to be generic with my approach for this so that the json/csv readers/writers can be used for other classes too.

Let me know what you think. Thanks!


vinopm:
@sam-brett @johnnonweiler Please can I have your review for this PR. Thanks

sam-brett (commented):
Happy with the test coverage.

Although, I'm slightly dubious about the Immutable library. I've used AutoValue and Lombok in the past and they haven't led to as readable/maintainable code as I'd hoped. I had leaned towards a POJO approach as I wanted to be able to pass the code back to Sibylle and Louise knowing it would be easy for them to go in and change in the future when RAMP wraps up. This is the reason I'd avoided Spring or any other DI frameworks as well. 

I guess it comes down to an overall design choice. As a developer I have no problems at all, and it's something I may look at using myself. This might be my preconceptions though. If @sibylleM and @johnnonweiler are happy then I am.

sibylleM:
If you think it's worthwhile to change to immutables then I'm happy with this. I've never used this myself but I've read that there are methods to create a new modified instance of the original object which do not modify the original object, but create a new instance with a modified field. If I understood correctly... So I guess there are ways to change the code.

vinopm:
Thanks for your feedback @sam-brett and @sibylleM. You are correct @sibylleM. The main advantage of using Immutables is that we can reduce a lot of unnecessary boilerplate without having to introduce much complexity. Immutables as a library is quite nice to use from a developers point of view. Its API is simple unlike Spring. To create a new object, you always use the fluent builder which is provided. For example instead of:
`StandardProperties properties = new StandardProperties(10000, 100, 1000, 0, true);`
we would use this:
```
StandardProperties properties = ImmutableStandardProperties.builder()
                .populationSize(10000)
                .timeLimit(100)
                .infected(1000)
                .seed(0)
                .steadyState(true)
                .build();
```
which I would argue is a lot more readable + easier on the developer to make sure they are passing in the correct values. It also ensures that we follow best practices regarding the importance of immutability (the objects state can not be changed, unless a copy of the object is made).

A lot of boilerplate code is also reduced, all we need to have is this as a definition:
```
@Immutable
public interface StandardProperties {
    int populationSize();
    int timeLimit();
    int infected();
    int seed();
    boolean steadyState();
}
```
instead of a full class with variable names, getters and setters.

I understand your concern to try and keep the implementation vanilla Java. From that point of view it makes sense not to introduce Spring, as it can be quite heavy-weight to pick up. However I believe that using Immutables where it makes sense to will be a good step towards a cleaner code base. Let me know if you agree/disagree. Happy to discuss further.

sam-brett:
I agree with the cleanliness of construction, especially if the IDE doesn't mark up or an order is changed. If Sibylle is happy to move towards this that allays my primary concern.

sibylleM:
Yep, happy with this. Thanks @vinopm and @sam-brett for your clarifications on this.

johnnonweiler (approved):
Looks good to me too!

I agree with pretty much all of the comments you have all been making.  I generally prefer to steer towards more standard idiomatic Java, but am also a fan of reducing boiler plate code.  In this case, it looks like a good change, and in particular the tests read really well.

(A couple of minor comments on details:
1) Why is `ExampleCsvPojo.java` in `src/main` - I thought it might fit better in `src/test`?
2) You added a "TODO" in SeirRecordTest.java, and I think it would be good to do it.)

johnnonweiler:
(Merging this seemed like the right thing to do.  Hope it was!)

PR #11, opened 2020-05-16
Sambrett/new schema

sam-brett:
Creating draft PR so I can resolve any conflicts


sam-brett:
@vinopm Hi Vino. I've been playing around with the immutable library a little. Could you have a look at the Person.java class. I'd like to set a default behaviour for the setStatus method so that ic calls the enum check transitionTo() method, but I don't seem to be able to work out a way for this to work. Maybe I just need to extend from the Immutable. Cheers
Sam

vinopm:
Hi @sam-brett , checking this for you now

vinopm:
@sam-brett I would suggest creating a new ImmutablePerson object with an updated status instead of using ModifiablePerson (there is limited support for this according to https://immutables.github.io/immutable.html#modifiable-classes). If Person objects are immutable, their inner state should not be modified once they are constructed. If we want to keep the object mutable, we should just use a plain Java POJO rather than enforce using Immutables here. 
However, I believe we can achieve a better design if we use enforce that the Person object is immutable, which will require some changes. I have created a draft PR into your newSchema branch: https://github.com/ScottishCovidResponse/Contact-Tracing-Model/pull/12. 
For the places where we intend to change the VirusStatus of the person, I am constructing a new person object and replacing the population map with the new person object. The checking of whether the transition is valid or not is done within the EvaluateCase class now.

I'd also like to suggest that we avoid introducing elements of randomness (through using `new Random()`) within the unit tests where it makes sense to do so, so that we do not have flaky tests

Once again, if you believe that the status (viral status) of a specific Person object needs to be modifiable after it is constructed, then we should probably avoid using Immutables here.

Let me know what you think!

sam-brett:
@vinopm I ended up taking a variation on that approach, I put the constant parts of a person (age gender etc) into an immutable called human, then made a class called Case which contains the ImmutableHuman and the statuses etc. This seems to be a reasonable compromise I think. 

Personally, I'm quite a fan of using randoms in tests because while they may expose a flaky test, they may also expose a missed edge case. I quite often do blast testing to highlight these conditions.

vinopm:
@sam-brett Nice, separating out to 'Case' looks good to me. 

Thanks for explaining, I agree that blast testing to highlight conditions which we may have missed is a great idea. Although I don't intend to nit pick, my 2c is that I think it could hurt us later on if the tests we write are not repeatable. For a given test/build run, we could skip running the test for a known input combination. If there was an issue with the code for this known input combination, the test run would not highlight it. Good unit tests should follow 'FIRST' - one of these acronyms point to 'repeatability': Tests should produce the same results each time & every time to ensure they are reliable and robust. https://howtodoinjava.com/best-practices/first-principles-for-good-tests/ 

I see the benefit of having blast testing, as it allows testing for cases we may not have considered. Maybe we could consider doing this as a separate suite of tests in addition to a standard set of unit tests, rather than a replacement for it?

src/main/java/uk/co/ramp/AppConfig.java
```diff
@@ -0,0 +1,80 @@
...
+            for (String beanName : beanNames) {
+                LOGGER.trace(beanName);
+            }
+        };
+    }
+
+
+    @Bean
+    public StandardProperties standardProperties() throws IOException {
+        try (Reader reader = new FileReader(new File("input/runSettings.json"))) {
```
vinopm: Move this: `input/runSettings.json` into a static constant string and declare it at top of the class as `STANDARD_PROPERTIES_LOCATION` rather than hardcoded within this method. Same for the other two below: diseaseSettings and populationSettings

src/main/java/uk/co/ramp/EvaluateCase.java
```diff
@@ -0,0 +1,225 @@
...
+            p.setNextVirusStatusChange(currentTime + timeToNextChange);
+        } else {
+            p.setNextVirusStatusChange(-1);
+        }
+
+        p.setStatus(newStatus);
+
+    }
+
+    public Set<Integer> checkTime(int time) {
```
vinopm: The method name is confusing here as the method is doing a lot more than checking if an actionable time has been reached.

src/main/java/uk/co/ramp/EvaluateCase.java
```diff
@@ -0,0 +1,225 @@
...
+        Set<Integer> alerts = new HashSet<>();
+        if (p.nextVirusStatusChange() == time) checkVirusStatus(time);
+        if (p.nextAlertStatusChange() == time) alerts = checkAlertStatus(time);
+
+        return alerts;
+
+
+    }
+
+    private void checkVirusStatus(int time) {
```
vinopm: The same over here, it should probably be called checkAndUpdateVirusStatus?

src/main/java/uk/co/ramp/EvaluateCase.java
```diff
@@ -0,0 +1,225 @@
...
+        switch (p.status()) {
+            case EXPOSED:
+                updateVirusStatus(EXPOSED_2, time);
+                break;
+            case EXPOSED_2:
+                VirusStatus status = determineInfection(p);
+                updateVirusStatus(status, time);
+                break;
+            case INFECTED:
+                updateVirusStatus(RECOVERED, time);
```
vinopm, line 100: Should there be some level of alerting followed by testing here as well? I know that currently there is no capacity for testing if there are no symptoms. However, in our model design, there is a mention of a proportion of individuals quarantining due to alerts followed by test for I<sub>A</sub>?

vinopm:
Hi @sam-brett  please see my previous comment, also the src/main/resources/contacts.csv file has been removed but not replaced with the new one?

src/main/java/uk/co/ramp/people/AlertStatus.java
```diff
@@ -0,0 +1,46 @@
...
+
+import java.util.Collections;
+import java.util.EnumMap;
+import java.util.List;
+import java.util.Map;
+
+public enum AlertStatus {
+
+    TESTED_POSITIVE,
+    TESTED,
```
vinopm: Does tested here mean that the results are negative?

sam-brett:
Morning @vinopm. My plan is to create a few test resources, including a fixed seed rng, so that'll give us the repeatability. I'd not heard the FIRST acronym before, I like the concept. I need to add tests for a lot of this, but I'm being pulled into some normal work today and a lot of tomorrow so I'll have to come back to that. 

With regards to the contacts file, these have been moved out of resources into the input folder, as I think they are going to be varied by the user. The input file is pointed to in the runSettings.

src/main/java/uk/co/ramp/people/AlertStatus.java
```diff
@@ -0,0 +1,46 @@
...
+    REQUESTED_TEST,
+    ALERTED,
```
vinopm, line 16: Is this in reference to being alerted from coming into contact with someone else who is infectious?

src/main/java/uk/co/ramp/people/AlertStatus.java
```diff
@@ -0,0 +1,46 @@
...
+    private static final Logger LOGGER = LogManager.getLogger(AlertStatus.class);
+    private static final Map<AlertStatus, List<AlertStatus>> validTransitions;
+
+    // Loaded as static to avoid forward reference error
+    static {
+        validTransitions = new EnumMap<>(AlertStatus.class);
+        validTransitions.put(NONE, List.of(ALERTED, REQUESTED_TEST));
+        validTransitions.put(ALERTED, List.of(REQUESTED_TEST));
+        validTransitions.put(REQUESTED_TEST, List.of(TESTED));
+        //TODO: get clarification here
```
vinopm, line 28: 👍

src/main/java/uk/co/ramp/people/Case.java
```diff
@@ -0,0 +1,113 @@
...
+    private int exposedTime;
+    private int nextVirusStatusChange;
+    private int nextAlertStatusChange;
+
+    public Case(final Human human) {
+        this.human = human;
+        status = SUSCEPTIBLE;
+        alertStatus = NONE;
+        contactRecords = new HashSet<>();
+        exposedBy = -9;
```
vinopm: magic number?

src/main/java/uk/co/ramp/ContactRunner.java
```diff
@@ -1,283 +1,74 @@
...
+            infection.setContactRecords(contactRecords);
+            infection.setPopulation(population);
 
-        LOGGER.info(infectedIds);
+            Map<Integer, CmptRecord> records = infection.propagate();
 
-        for (Integer id : infectedIds) {
-            population.get(id).updateStatus(INFECTED, 0);
-            LOGGER.info("population.get(id).getNextStatusChange() = {}", population.get(id).getNextStatusChange());
+            writeCompartments(new ArrayList<>(records.values()), new File("Compartments.csv"));
```
vinopm: Move location string as static declaration at top or make this configurable

src/test/java/uk/co/ramp/people/PopulationGeneratorTest.java
```diff
@@ -257,19 +261,20 @@ public void testGeneratePopulation() {
...
 
 
-    private void createPeople(Map<Integer, Person> var, int start, int end, VirusStatus v) {
+    private void createPeople(Map<Integer, Case> var, int start, int end, VirusStatus v) {
         for (int p = start; p < end; p++) {
 
-            Person person = Mockito.mock(Person.class);
-            when(person.getStatus()).thenReturn(v);
-            var.put(p, person);
+//            Human person = Mockito.mock(Human.class);
```
vinopm: remove comment

src/test/java/uk/co/ramp/people/PopulationGeneratorTest.java
```diff
@@ -257,19 +261,20 @@ public void testGeneratePopulation() {
 
         Assert.assertEquals(populationSize, men + women);
 
-        Assert.assertEquals(0.5, men / (double) populationSize, 1d / Math.sqrt(populationSize));
-        Assert.assertEquals(0.5, women / (double) populationSize, 1d / Math.sqrt(populationSize));
+        Assert.assertEquals(0.5, men / (double) populationSize, 0.15);
```
vinopm: The delta of 0.15 is very large here. Is this because we are using a random input set, and we expect the range to be +/- 0.15? Are we guaranteed to hit the +/- 0.15 range through every single run of the test?

src/main/java/uk/co/ramp/people/PopulationGenerator.java
```diff
@@ -1,39 +1,46 @@
...
-        Map<Integer, Person> population = new HashMap<>();
-        for (int i = 0; i < runProperties.populationSize(); i++) {
-
-            int age = findAge(r.nextUniform(0, 1), cumulative, properties.populationAges());
-            Gender g = r.nextUniform(0, 1) > properties.genderBalance() / 2d ? Gender.FEMALE : Gender.MALE;
-            double compliance = r.nextGaussian(0.5, 0.5);
-            double health = r.nextGaussian(0.5, 0.5);
-
-            population.put(i, new Person(i, age, g, compliance, health));
+    public static Map<VirusStatus, Integer> getCmptCounts(Map<Integer, Case> population) {
```
vinopm, line 35: Could use streams here instead:

```
return population.values().stream()
                .map(Case::status)
                .collect(Collectors.groupingBy(Function.identity(), Collectors.summingInt(e -> 1)));
```

src/main/java/uk/co/ramp/people/PopulationGenerator.java
```diff
@@ -87,39 +94,28 @@ int findAge(double v, Map<Integer, Double> c, Map<Integer, MinMax> populationAge
         return cumulative;
```
vinopm, line 84: Question regarding this block:
```
if (i == data.size() - 1) {
     sum = 1d; 
}	            
```
Has this been put in to deal with cases where the population distribution input does not add up to 1 in the json input file? If this is the case shouldn't we scale up the distribution sum to 1 rather than artificially set the last bin (age = 4) to have a higher proportion.

src/main/java/uk/co/ramp/people/PopulationGenerator.java
```diff
@@ -1,39 +1,46 @@
...
+
+        for (Case p : population.values()) {
+            counts.merge(p.status(), 1, Integer::sum);
         }
 
-        return population;
+        return counts;
     }
 
     int findAge(double v, Map<Integer, Double> c, Map<Integer, MinMax> populationAges) {
```
vinopm: We could replace this hand-rolled implementation of interpolation to find the sample age with the integer distribution library provided by Apache Commons Math API (which we already use as a dependency within the project). 


Here is a suggestion, where I pass the discrete non-cumulative distribution into the API. Using this allows us to remove the `createCumulative()` method as well as we would not need to pass in a cumulative distribution anymore:

```
import org.apache.commons.math3.distribution.EnumeratedIntegerDistribution;

int findAge(Map<Integer, Double> populationDistribution, Map<Integer, MinMax> populationAges) {
    int maxAge = populationAges.values().stream().mapToInt(MinMax::max).max().orElseThrow();
    int[] outcomes = IntStream.rangeClosed(0, maxAge).toArray();
    double[] probabilities = IntStream
            .range(0, populationAges.size())
            .mapToObj(idx -> IntStream
                    .rangeClosed(populationAges.get(idx).min(), populationAges.get(idx).max())
                    .mapToDouble(age -> populationDistribution.get(idx) / (populationAges.get(idx).max() - populationAges.get(idx).min() + 1))
                    .toArray())
            .flatMapToDouble(DoubleStream::of)
            .toArray();

    EnumeratedIntegerDistribution distribution = new EnumeratedIntegerDistribution(dataGenerator.getRandomGenerator(), outcomes, probabilities);
    return distribution.sample();
}
```
This can then easily allow for the user to create more bins in the json data if they want to, without having to change the code

src/main/java/uk/co/ramp/EvaluateCase.java
```diff
@@ -0,0 +1,225 @@
...
+import uk.co.ramp.people.InvalidStatusTransitionException;
+import uk.co.ramp.people.VirusStatus;
+
+import java.util.HashSet;
+import java.util.Set;
+
+import static uk.co.ramp.people.AlertStatus.*;
+import static uk.co.ramp.people.VirusStatus.*;
+
+public class EvaluateCase {
```
vinopm, line 20: Missing unit tests for this class I think?

src/main/java/uk/co/ramp/Outbreak.java
```diff
@@ -0,0 +1,339 @@
...
+import java.io.IOException;
+import java.io.Writer;
+import java.util.*;
+
+import static uk.co.ramp.people.AlertStatus.ALERTED;
+import static uk.co.ramp.people.AlertStatus.NONE;
+import static uk.co.ramp.people.VirusStatus.*;
+
+@Service
+public class Outbreak {
```
vinopm, line 28: Missing unit tests for this class as well I think?

src/main/java/uk/co/ramp/EvaluateCase.java
```diff
@@ -0,0 +1,225 @@
...
+        switch (p.alertStatus()) {
+
+            case TESTED_POSITIVE:
+                LOGGER.trace("user {} has tested positive", p.id());
+                return alertAllContacts();
+            case TESTED:
+                // TODO: maybe include flag for has had virus?
+                // TODO: should a recovered person test +ve?
+                if (p.isInfectious() || p.status() == DEAD) {
+                    LOGGER.trace("user {} has tested positive", p.id());
```
vinopm, line 128: Is there a reason for the distinction between this case and TESTED_POSITIVE?

src/main/java/uk/co/ramp/Outbreak.java
```diff
@@ -0,0 +1,339 @@
...
+        Set<Integer> infectedIds = infectPopulation();
+        for (Integer id : infectedIds) {
+
+            EvaluateCase evaluateCase = new EvaluateCase(population.get(id), diseaseProperties, rng);
+            evaluateCase.updateVirusStatus(EXPOSED, 0);
+
+        }
+    }
+
+    int tab = 0;
```
vinopm: move to top and mark as private

src/main/java/uk/co/ramp/Outbreak.java
```diff
@@ -0,0 +1,339 @@
...
+        // randomly infected
+        infectees.addAll(infectors.getOrDefault(-2, Collections.emptySet()));
+
+        List<Integer> target = new ArrayList<>(infectees);
+
+        try (Writer writer = new FileWriter(new File("infectionMap.txt"))) {
+            recurseSet(target, infectors, writer);
+        } catch (IOException e) {
+            LOGGER.error(e.getMessage());
+            // TODO: handle properly
```
vinopm: TODO to be resolved

src/main/java/uk/co/ramp/Outbreak.java
```diff
@@ -0,0 +1,339 @@
...
+            infectors.putIfAbsent(c.exposedBy(), new HashSet<>());
+
+            Set<Integer> var = infectors.get(c.exposedBy());
+
+            var.add(c.id());
+            infectors.put(c.exposedBy(), var);
+        }
+
+        // initial infectees
+        Set<Integer> infectees = infectors.get(-9);
```
vinopm: -9 is this a magic number? If it is can we declare it as a static and move it to top as a class variable?

src/main/java/uk/co/ramp/Outbreak.java
```diff
@@ -0,0 +1,339 @@
...
+
+        // randomly infected
+        infectees.addAll(infectors.getOrDefault(-2, Collections.emptySet()));
+
+        List<Integer> target = new ArrayList<>(infectees);
+
+        try (Writer writer = new FileWriter(new File("infectionMap.txt"))) {
```
vinopm: move file location/name to be a class variable or static for now

vinopm (changes requested):

vinopm:
@sam-brett I have reviewed your PR. Thanks for your response to the randomized testing. It is true that with a fixed seed, we can ensure repeatability, but it doesn't guarantee that we will run tests for all known inputs that we know we should explicitly test for. In this case we would be consistently missing out testing a subset of known input parameters.

src/main/java/uk/co/ramp/EvaluateCase.java
sam-brett: Renamed to checkActionsAtTimestep

src/main/java/uk/co/ramp/people/AlertStatus.java
sam-brett: Changed TESTED to AWAITING_RESULT

src/main/java/uk/co/ramp/people/AlertStatus.java
sam-brett: Yes

src/main/java/uk/co/ramp/people/Case.java
sam-brett: Added     :
private static final int DEFAULT = -9;
private static final int RANDOM_INFECTION = -2;

src/main/java/uk/co/ramp/people/PopulationGenerator.java
sam-brett: It was just to tidy up any rounding errors when adding doubles. The current input adds to 1 in the real world, but 0.9999 in binary.

src/main/java/uk/co/ramp/people/PopulationGenerator.java
sam-brett: That's great.

src/main/java/uk/co/ramp/people/PopulationGenerator.java
sam-brett: Very tidy

sam-brett: This doesn't include empty entries for places where the sum is 0. This means we'd need to add a lot of getOrDefaults elsewhere.

vinopm: You could add one extra line below for this:

`Stream.of(VirusStatus.values()).forEach(vs -> res.putIfAbsent(vs, 0));`
where `res` refers to the map created in my first comment.

sam-brett: Added extra line as suggested.

src/main/java/uk/co/ramp/Outbreak.java
sam-brett: Done

src/main/java/uk/co/ramp/Outbreak.java
sam-brett: Yeah, I'm a bit behind with the testing, but I'm going to spawn that off as a new ticket.

src/main/java/uk/co/ramp/AppConfig.java
sam-brett: Done

src/test/java/uk/co/ramp/people/PopulationGeneratorTest.java
sam-brett: Changed back to 1/rt(n)

src/test/java/uk/co/ramp/people/PopulationGeneratorTest.java
sam-brett: Done

src/main/java/uk/co/ramp/people/PopulationGenerator.java
```diff
@@ -100,7 +90,7 @@ int findAge(double v, Map<Integer, Double> c, Map<Integer, MinMax> populationAge
 
         for (int i = 0; i < runProperties.populationSize(); i++) {
 
-            int age = findAge(dataGenerator.nextUniform(0, 1), cumulative, properties.populationAges());
+            int age = findAge(cumulative, properties.populationAges());
```
vinopm: @sam-brett We should pass in the non-cumulative distribution: `properties.populationDistribution()` here instead of `cumulative`

src/main/java/uk/co/ramp/EvaluateCase.java
vinopm: A question for @sibylleM:
In the diagram we have for the proposed model, for I<sub>A</sub> it states that a proportion of these people will go through testing and then quarantine if positive. Currently in our implementation, anyone with status of I<sub>A</sub> will not go through testing. Is this the intended/correct behaviour?

sam-brett: When someone is asymptomatically infected, they won't know to quarantine unless they are alerted that they've been in contact with someone who is a confirmed case (the root case will be someone who is symptomatic). Once they are alerted they are able to get tested.  The I<sub>a</sub> group will still test positive when tested, after which their contacts can be alerted.

src/main/java/uk/co/ramp/people/PopulationGenerator.java
vinopm: Also, strange that the unit test which tests this method is not failing here.

src/main/java/uk/co/ramp/EvaluateCase.java
```diff
@@ -0,0 +1,230 @@
...
+    int getDistributionValue(double mean, double max) {
+
+        int value = (int) Math.round(mean);
+        double sample;
+        switch (properties.progressionDistribution()) {
+            case GAUSSIAN:
+                sample = rng.nextGaussian(mean, mean / 2d);
+                break;
+            case LINEAR:
+                sample = rng.nextUniform(mean - mean / 2d, mean + mean / 2d);
```
vinopm, line 214: Hi @sam-brett , can you explain why we are generating a sample between `0.5*mean` and `1.5*mean` for LINEAR instead of from `mean - (max - mean)` to `max`?

sam-brett: That’ll just be a hangover from when the mean was the only variable.

src/main/java/uk/co/ramp/EvaluateCase.java
```diff
@@ -0,0 +1,230 @@
...
+
+        LOGGER.trace("Alerting " + contactIds);
+        return contactIds;
+
+    }
+
+
+    private void updateAlertStatus(final AlertStatus newStatus, final int currentTime) {
+
+        int time = getDistributionValue(properties.meanTestTime(), properties.maxTestTime());
```
vinopm, line 174: @sam-brett We are using the DiseaseProperties distribution which is exponential here. Should we be using a linear distribution here or is this intentional? Thanks

src/main/java/uk/co/ramp/EvaluateCase.java
```diff
@@ -0,0 +1,230 @@
...
+                return alertAllContacts();
+            case AWAITING_RESULT:
+                // TODO: maybe include flag for has had virus?
+                // TODO: should a recovered person test +ve?
+                if (p.wasInfectious()) {
+                    LOGGER.trace("user {} has tested positive", p.id());
+                    updateAlertStatus(TESTED_POSITIVE, time);
+                } else {
+                    LOGGER.trace("user {} has tested negative", p.id());
+                    updateAlertStatus(NONE, time);
```
vinopm, line 132: @sam-brett  Question about this line which is confusing me a bit: It makes sense to me why we set the alert status to NONE here, however in updateAlertStatus() we then go and set the nextAlertStatusChange time to some point in the future. I believe this could then cause the alertstatus to change from NONE to ALERTED again?

src/main/java/uk/co/ramp/EvaluateCase.java
```diff
@@ -0,0 +1,230 @@
...
+                sample = rng.nextExponential(mean);
+                break;
+            case FLAT:
+            default:
+                return value;
+        }
+
+        value = (int) Math.round(sample);
+
+        return Math.min(Math.max(value, 1), (int) max);
```
vinopm: For the exp distribution, setting values which are greater than max to artificially become equal to max causes a large spike in the frequency of value=max being returned. This doesn’t go inline with the exp distribution where the frequency of max should be very low. Some of the other distributions will also have the same issue. We could get around this by recursively calling the method again when this case ( value > max or value < 1) is detected. This would ensure the properties of the distribution are still held whilst not returning back any values > max or < 1. Just a suggestion, maybe there is a better way to resolve this? Happy to hear suggestions or discuss further.

src/main/java/uk/co/ramp/EvaluateCase.java
vinopm: Question about this line which is confusing me a bit: It makes sense to me why we set the alert status to NONE here, however in updateAlertStatus() we then go and set the nextAlertStatusChange time to some point in the future. I believe this could then cause the alertstatus to change from NONE to ALERTED again?

src/main/java/uk/co/ramp/EvaluateCase.java
vinopm, line 174: We are sharing the DiseaseProperties distribution which is exponential here. Should we be using a linear distribution here or is this intentional? Thanks

vinopm (changes requested):

sam-brett:
Hey @vinopm, sorry I was in the office most of yesterday so I'm only getting around to these now. I'll comment on each issue individually, but hopefully we can get this merged today.

src/main/java/uk/co/ramp/people/PopulationGenerator.java
sam-brett: Will change.

src/main/java/uk/co/ramp/EvaluateCase.java
sam-brett: That's a very good point. That sounds like a sound resolution too, I'll implement this. 

See before and after:

Screenshot 2020-05-27 at 10 52 07

src/main/java/uk/co/ramp/EvaluateCase.java
sam-brett: I think you're right, I've been debating this whole area and I'm going to set up an event list to avoid all this logic being as part of a case. There are some quirks that I need to speak to Louise and Sibylle about as well, because we have the wasInfectious field, which tests saves if the person was infected when the test was taken, but they can then go on to die with a test being negative.

src/main/java/uk/co/ramp/EvaluateCase.java
sam-brett: This is intentional. I've included some fields so they can sample the time it takes for tests and alerts to occur. Again this needs some ironing out.

sam-brett (commented):
I think thats all the comments addressed.

src/main/java/uk/co/ramp/EvaluateCase.java
vinopm: Ok thanks for clarifying

vinopm (approved):

PR #13, opened 2020-05-28
Adding documentation

sam-brett:
Please find the first draft of the document in the /docs directory of the repo. 

If anyone has a better suggestion I'm open to it, but it might be best to create a copy with track changes and have them all merged by a single user.


vinopm:
@sam-brett Could we have this converted to markdown file format, so that it is viewable on Github. I'll read through the documentation content today. Nice work!

johnnonweiler:
@vinopm Much as having documentation in markdown format is generally a good idea, I think it would be awkward to put this particular document into markdown format.  And it's not very difficult to download and view single Word documents from GitHub.

(LaTeX is another option, but there are many people who are much more comfortable with MS Word, so I think that on balance a Word document is probably the way to go.)

vinopm:
@johnnonweiler sure, however it will be difficult to PR review the lines which have changed in the document if it is in docx format. That said, happy to keep it as docx.

sam-brett:
@johnnonweiler @vinopm The MD idea is a good one, I guess it we can discuss that at the meeting tomorrow as I'm assuming Louise will "own" the document and she may want it as more of an offline thing.

sibylleM:
@johnnonweiler @vinopm @sam-brett I think it's a good idea to add a markdown version and makes it easy to add to websites etc.. How about I'll produce an additional .md file based on Sam's documentation?

sibylleM:
Hi @sam-brett @johnnonweiler @vinopm, .md file of the documentation can be found in the /docs directory of the repo. Let me know if anything looks weird. Otherwise we can close this issue.

sibylleM:
Hi @sam-brett @vinopm @johnnonweiler,  Took me a while to figure out how to render the equations etc. The new version can be found in Contact-Tracing-Model/docs/doc_v1.0/Build_documentation.md. I had to save the images and equations in the same folder, otherwise it does not display correctly. Looks good to me but please let me know if you spot anything weird.

PR #15, opened 2020-06-01
Initial version of isolation policies

vinopm:
PR for https://github.com/ScottishCovidResponse/SCRCIssueTracking/issues/259

**Stage 1**

- [x] create a new input file to contain policy data 
- [x] set a proportion of the population infected where lockdown is imposed
- [x] set a proportion where the lockdown is relaxed
- [ ] Set a flag for whether isolation is relative (6 in above example) or absolute (14 in above example)
- [x] set length of absolute isolation

**Stage 2**

- [x] update existing code to use these new features
- [x] include todo's wherever input from Sibylle or Louise are required
- [x] use best judgement but leave TODO's where unsure

Some more notes:
- I have implemented this with the aim of keeping this generic and flexible, so that isolation policies can be added/removed/amended easily without requiring any code changes.
- Isolation policies can be specified via json in  `input/isolationPolicies.json`
- For this initial version, the start of isolation time is considered to be the time when an individual first starts to isolate.
- An individual can isolate multiple times if model requires individuals to do so. i.e. if they are alerted and test negative and then become infected & symptomatic later on.
- Isolation policies are split into three types:
1. Global policies
- Input parameters specify when to trigger policy guidelines. For example:
```
"Stay at Home Policy": {
      "proportionInfected": {
        "min": 20,
        "max": 100
      },
      "isolationProperty": {
        "isolationProbabilityDistribution": {
          "type": "FLAT",
          "mean": 90,
          "max": 90
        },
        "priority": 1
      }
    }
```
This policy will be activated when the % infected is > 20%. The isolationProperty determines the action of the policy. In this case, we will provide guidance to 90% of people that they should isolate.

Multiple global policies can be specified, and it is possible to have an overlap of condition. In this case, the priority is used to determine which policy to invoke. We consider highest priority to be the priority with highest value i.e. 1 has higher priority than 0.

2. Individual policies (Used for determining what to do on a change of VirusStatus & AlertStatus of an individual)

An example: 

```
"Infected Symptomatic Policy": {
        "virusStatus": "INFECTED_SYMP",
        "isolationProperty": {
          "isolationProbabilityDistribution": {
            "type": "FLAT",
            "mean": 100,
            "max": 100
          },
          "isolationTimeDistribution": {
            "type": "LINEAR",
            "mean": 13,
            "max": 14
          },
          "priority": 1
        }
      }
```
This policy will trigger when an individual is infected and symptomatic. It will advise 100% of individuals in this state to isolate for a period of 12-14 days.

Also it is important to note that individuals can not comply to the isolation guidelines, this is dependent on the compliance value for each person. (already previously implemented).

3. Default policy
If there are no applicable policies for the individual, we fallback to the default policy:
```
"isolationProbabilityDistribution": {
      "type": "FLAT",
      "mean": 0,
      "max": 0
    },
    "priority": 0
  }
```
which conditions to no isolation. Any cases which are not covered in other rules will default to this one. An example of this is when someone is tested and result is negative. As there is no explicit policy for this, the action will use the default policy, which will result in no isolation. In this case, if a person was previously set to isolate for 14 days and they were tested negative on day 5, they are not told to isolate for the remainder of days.


vinopm:
@sam-brett I'll address "Set a flag for whether isolation is relative (6 in above example) or absolute (14 in above example)" in a separate PR to avoid too much complexity in this PR. I believe we should do this after the 'event list' mechanism is put into place.

src/main/java/uk/co/ramp/policy/ContactIsolationPolicy.java
```diff
@@ -0,0 +1,32 @@
...
+
+    ContactIsolationPolicy(SingleCaseIsolationPolicy singleCaseIsolationPolicy, DiseaseProperties diseaseProperties) {
+        this.singleCaseIsolationPolicy = singleCaseIsolationPolicy;
+        this.diseaseProperties = diseaseProperties;
+    }
+
+    @Override
+    public boolean isContactIsolated(Case caseA, Case caseB, double contactWeight, double actualInfectedProportion, int currentTime) {
+        boolean isCaseAInIsolation = singleCaseIsolationPolicy.isIndividualInIsolation(caseA.id(), caseA.status(), caseA.alertStatus(), caseA.compliance(), actualInfectedProportion, currentTime);
+        boolean isCaseBInIsolation = singleCaseIsolationPolicy.isIndividualInIsolation(caseB.id(), caseA.status(), caseA.alertStatus(), caseA.compliance(), actualInfectedProportion, currentTime);
```
sam-brett: should this line reference caseA? or should it be:
`
isIndividualInIsolation(caseB.id(), caseB.status(), caseB.alertStatus(), caseB.compliance(), actualInfectedProportion, currentTime);
`

if so, it might be cleaner to send the caseA and caseB objects here and get the required properties in isIndividualInIsolation(). I might be misinterpreting though, in which case a comment will do.

src/main/java/uk/co/ramp/policy/SingleCaseIsolationPolicy.java
```diff
@@ -0,0 +1,93 @@
...
+                .filter(policy -> policy.virusStatus() == virusStatus)
+                .map(VirusStatusIsolationProperty::isolationProperty);
+
+        Stream<IsolationProperty> alertIsolationPolicy = isolationProperties.individualIsolationPolicies().alertStatusPolicies().values().stream()
+                .filter(policy -> policy.alertStatus() == alertStatus)
+                .map(AlertStatusIsolationProperty::isolationProperty);
+
+        Stream<IsolationProperty> defaultIsolationPolicy = Stream.of(isolationProperties.defaultPolicy());
+
+        return Stream.of(globalIsolationPolicy, virusIsolationPolicy, alertIsolationPolicy, defaultIsolationPolicy)
```
sam-brett: As this is relying on user inputs, it might be worth adding a check to ensure that policies don't have the same priority so we don't get a clash here.

src/main/java/uk/co/ramp/policy/VirusStatusIsolationProperty.java
```diff
@@ -0,0 +1,12 @@
+package uk.co.ramp.policy;
+
+import org.immutables.gson.Gson;
+import org.immutables.value.Value;
+import uk.co.ramp.people.VirusStatus;
+
```
sam-brett, line 6: It might be good to move some of the immutable objects into a sub-package to clear up the policy level a little.

src/main/java/uk/co/ramp/policy/SingleCaseIsolationPolicy.java
sam-brett: @sibylleM As the FMD model was in Java 7, have you used streams much? If so we can provide comments as descriptions.

src/main/java/uk/co/ramp/policy/ContactIsolationPolicy.java
vinopm: You're right. Copy paste error . I'll make the change as you suggested and additional unit tests

src/main/java/uk/co/ramp/distribution/DistributionSampler.java
```diff
@@ -0,0 +1,51 @@
+package uk.co.ramp.distribution;
+
+import org.apache.commons.math3.random.RandomDataGenerator;
+
+public class DistributionSampler {
+    private final RandomDataGenerator rng;
```
sam-brett, line 6: I had this on my todo list, so thanks for breaking this out.

src/main/java/uk/co/ramp/policy/SingleCaseIsolationPolicy.java
vinopm: Sure, I'll add in a check

src/main/java/uk/co/ramp/policy/ContactIsolationPolicy.java
sam-brett: I couldn't make out if I was being dense or not.

src/main/java/uk/co/ramp/Framework.java
```diff
@@ -14,7 +15,7 @@
```
sam-brett, line 15: I've made a bit of a clean up here in the event list task.

sam-brett (changes requested):
Very minor changes, but looks good.

src/main/java/uk/co/ramp/policy/ContactIsolationPolicy.java
vinopm: It's a good spot!

src/main/java/uk/co/ramp/policy/VirusStatusIsolationProperty.java
vinopm: I prefer to structure new features (such as the isolation policy) together in their own individual packages. 
Some more info here on the concept of 'package by feature, not layer': http://www.javapractices.com/topic/TopicAction.do?Id=205#:~:text=The%20package%2Dby%2Dfeature%20style%20still%20honors%20the%20idea%20of,t%20seem%20necessary%20or%20desirable.

This allows us to make everything within the 'policy' package be either 'package private' or 'private' with the exception of the isContactIsolated() method in the IsolationPolicy interface which will be 'public'. This ensures that all implementation details of the isolation policy are encapsulated within the policy package, and not exposed unnecessarily to other packages.

Please let me know if I'm missing your point

sam-brett: That's fair enough. I've added that to my reading list. It was a style thing, but that's ver person anyway.

vinopm:
@sam-brett updated PR with your suggestions

vinopm:
@sam-brett updated PR with your suggestions. Please review/approve, thanks

sam-brett (approved):
Looks good.

PR #16, opened 2020-06-04
Import initial cases

sam-brett:
This isn't a very big change, but allow Sibylle to test contacts with super spreaders etc.


src/main/java/uk/co/ramp/ContactRunner.java
```diff
@@ -25,13 +26,10 @@
 
     private static final Logger LOGGER = LogManager.getLogger(ContactRunner.class);
     public static final String COMPARTMENTS_CSV = "Compartments.csv";
-
+    private InputFiles inputFiles;
```
vinopm, line 29: Might be better to make it explicit that `InputFiles` retrieves the file location rather than the actual settings/data

```suggestion
        try (Reader reader = new FileReader(inputFileLocation.contactData())) {
```

vinopm: Might be better to make it explicit that `InputFiles` retrieves the file location rather than the actual settings/data

```suggestion
        try (Reader reader = new FileReader(inputFileLocation.contactData())) {
```

vinopm (commented):
Quite difficult to see the changes on this PR as it includes changes from the other PR which hasn't been merged yet into develop. I am looking at https://github.com/ScottishCovidResponse/Contact-Tracing-Model/compare/sambrett/unittests...importInitialCases which shows me a difference between the two branches, but not 100% sure if that's the right diff to look at. @sam-brett thoughts?

sam-brett:
Hey Vino, yeah that's the right change set. This will need to be merge post the unit test merge.

vinopm (approved):

PR #17, opened 2020-06-04
Event list

sam-brett:
Still adding tests, but the code can be reviewed as it is pretty static


sam-brett:
@vinopm I've made a fairly dirty merge here, but its allows it to merge automatically now. There is one test that has broken in the Isolation policy area (testShouldIsolate_UsesIndividualPolicy_InfectedAndNoneTestedPositive) I think this might be to do with the infection schema change.

src/main/java/uk/co/ramp/event/types/FormattedEvent.java
```diff
@@ -0,0 +1,25 @@
...
+
+import com.fasterxml.jackson.annotation.JsonPropertyOrder;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import org.immutables.value.Value;
+
+@Value.Immutable
+@JsonSerialize
+@JsonDeserialize
+@JsonPropertyOrder({"time", "eventType", "id", "newStatus", "additional"})
```
vinopm:
```suggestion
@JsonPropertyOrder({"time", "eventType", "id", "newStatus", "additionalInfo"})
```

src/main/java/uk/co/ramp/io/ContactReader.java
```diff
@@ -0,0 +1,41 @@
...
+import java.io.IOException;
+import java.io.Reader;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+@Service
+public class ContactReader {
+    private static final Logger LOGGER = LogManager.getLogger(ContactReader.class);
```
vinopm: Unused: 
```suggestion
```

src/main/java/uk/co/ramp/io/ContactReader.java
```diff
@@ -0,0 +1,41 @@
...
+
+    public Map<Integer, List<ContactEvent>> readEvents(Reader reader, StandardProperties properties) throws IOException {
```
vinopm:
```suggestion
    private final StandardProperties properties;
    @Autowired
    public ContactReader(StandardProperties standardProperties) {
        this.properties = standardProperties;
    }

    public Map<Integer, List<ContactEvent>> readEvents(Reader reader) throws IOException {
```
We could avoid passing in standardproperties (which is pretty much static, does not change per method call) by passing it through constructor and injecting it via Spring

src/main/java/uk/co/ramp/io/ContactReader.java
```diff
@@ -0,0 +1,41 @@
...
+            if (event.from() >= properties.populationSize() ||
+                    event.to() >= properties.populationSize() ||
+                    event.time() > properties.timeLimit()) continue;
+
+            dailyRecord.putIfAbsent(event.time(), new ArrayList<>());
+            dailyRecord.get(event.time()).add(event);
+
+        }
+
+        return dailyRecord;
```
vinopm:
```suggestion
return contactEvents.stream()
                .filter(event -> event.from() < properties.populationSize())
                .filter(event -> event.to() < properties.populationSize())
                .filter(event -> event.time() <= properties.timeLimit())
                .collect(Collectors.groupingBy(ContactEvent::time));
```

src/main/java/uk/co/ramp/people/AlertStatus.java
```diff
@@ -26,7 +27,8 @@
         validTransitions.put(ALERTED, List.of(REQUESTED_TEST));
         validTransitions.put(REQUESTED_TEST, List.of(AWAITING_RESULT));
         //TODO: get clarification here
-        validTransitions.put(AWAITING_RESULT, List.of(TESTED_POSITIVE, ALERTED, NONE));
+        validTransitions.put(AWAITING_RESULT, List.of(TESTED_POSITIVE, TESTED_NEGATIVE));
+        validTransitions.put(TESTED_NEGATIVE, List.of(NONE));
         validTransitions.put(TESTED_POSITIVE, Collections.emptyList());
     }
```
vinopm, line 34: transitionTo() not used anymore. we can remove this?

src/main/java/uk/co/ramp/people/Case.java
```diff
@@ -1,156 +1,133 @@
...
 
-    public void setStatus(VirusStatus newStatus) {
-        this.status = this.status.transitionTo(newStatus);
-    }
+    //derived values
 
-    public int nextAlertStatusChange() {
-        return nextAlertStatusChange;
+    public String getSource() {
+        return Strings.padEnd(id() + "(" + exposedTime + ")", 12, ' ');
```
vinopm: The padding should probably live within the implementation of InfectionMap rather than Case? id and exposedTime can already be accessed from InfectionMap

src/main/java/uk/co/ramp/people/Case.java
```diff
@@ -1,156 +1,133 @@
...
-    // statics
+        if (event instanceof VirusEvent) {
+            processVirusEvent((VirusEvent) event);
+        } else if (event instanceof InfectionEvent) {
+            processInfectionEvent((InfectionEvent) event);
+        } else if (event instanceof AlertEvent) {
+            processAlertEvent((AlertEvent) event);
+        } else {
+            throw new EventException("uncovered condition");
+        }
```
vinopm: Relying on `instanceof` instead of polymorphism is generally considered code smell within OOP. We could possibly resolve this by having an interface method in Event called something like applyEventToCase(Case aCase) and the children of Event such as VirusEvent/InfectionEvent/AlertEvent to have their own implementations for applyEventToCase.

For example the VirusEvent interface would have:
```
default void applyEventToCase(Case aCase) {
    aCase.setVirusStatus(aCase.getVirusStatus().transitionTo(newStatus());
}
```
or 
```
public void applyEventToCase(Case aCase) {
    aCase.setVirusStatus(aCase.getVirusStatus().transitionTo(newStatus());
}
```

and then in EventProcessor you would have 
`event.applyEventToCase(population.get(event.id())` instead of `population.get(event.id()).processEvent(event)`

src/test/java/uk/co/ramp/policy/SingleCaseIsolationPolicyTest.java
```diff
@@ -132,22 +128,24 @@ public void setUp() {
...
-        assertThat(isolationPolicy.isIndividualInIsolation(id, INFECTED_SYMP, TESTED_POSITIVE, compliance, proportionOfPopulationInfected, currentTime)).isTrue();
+        assertThat(isolationPolicy.isIndividualInIsolation(id, SYMPTOMATIC, TESTED_POSITIVE, compliance, proportionOfPopulationInfected, currentTime)).isTrue();
     }
 
     @Test
+    @Ignore
+    //TODO fix
     public void testShouldIsolate_UsesIndividualPolicy_InfectedAndNoneTestedPositive() {
-        assertThat(isolationPolicy.isIndividualInIsolation(id, INFECTED, NONE, compliance, proportionOfPopulationInfected, currentTime)).isFalse();
+        assertThat(isolationPolicy.isIndividualInIsolation(id, SYMPTOMATIC, NONE, compliance, proportionOfPopulationInfected, currentTime)).isFalse();
```
vinopm:
```suggestion
        assertThat(isolationPolicy.isIndividualInIsolation(id, ASYMPTOMATIC, NONE, compliance, proportionOfPopulationInfected, currentTime)).isFalse();
```

This should be ASYMPTOMATIC instead. Perhaps the test name should be updated too: testShouldIsolate_UsesIndividualPolicy_AsymptomaticAndNoAlertStatus.

src/test/java/uk/co/ramp/policy/SingleCaseIsolationPolicyTest.java
```diff
@@ -132,22 +128,24 @@ public void setUp() {
 
     @Test
     public void testShouldIsolate_UsesIndividualPolicy_InfectedAndTestedPositive() {
-        assertThat(isolationPolicy.isIndividualInIsolation(id, INFECTED_SYMP, TESTED_POSITIVE, compliance, proportionOfPopulationInfected, currentTime)).isTrue();
+        assertThat(isolationPolicy.isIndividualInIsolation(id, SYMPTOMATIC, TESTED_POSITIVE, compliance, proportionOfPopulationInfected, currentTime)).isTrue();
     }
 
     @Test
+    @Ignore
+    //TODO fix
```
vinopm:
```suggestion

```

src/main/java/uk/co/ramp/event/EventList.java
```diff
@@ -0,0 +1,83 @@
...
+    public void addEvents(Map<Integer, List<ContactEvent>> readEvents) {
+
+        for (Map.Entry<Integer, List<ContactEvent>> entry : readEvents.entrySet()) {
+            bulkAdd(entry.getKey(), entry.getValue());
+        }
+    }
+
+    private void bulkAdd(int time, List<ContactEvent> events) {
+        map.computeIfAbsent(time, k -> new ArrayList<>()).addAll(events);
+    }
```
vinopm:
```suggestion
    public void addEvents(Map<Integer, List<ContactEvent>> readEvents) {
        readEvents.values().stream().flatMap(List::stream).forEach(this::addEvent);
    }
```

src/main/java/uk/co/ramp/event/EventList.java
```diff
@@ -0,0 +1,83 @@
...
+
+
+    //read
+    public List<Event> getForTime(int time) {
+        return map.getOrDefault(time, new ArrayList<>());
+    }
+
+    public Map<Integer, List<Event>> getMap() {
+        return map;
+    }
```
vinopm: It is unnecessary to expose the map publicly. The only caller of this is from outbreak where outbreak.java is computing the last contact. We can replace this method with the following:
```suggestion
    public Optional<Integer> lastContactTime() {
        return map.keySet().stream().max(Comparator.naturalOrder());
    }
```

and in Outbreak.java we can call:
`int lastContact = eventList.lastContactTime().orElseThrow();`
instead of 
`int lastContact = eventList.getMap().keySet().stream().max(Comparator.naturalOrder()).orElseThrow();`

src/main/java/uk/co/ramp/event/EventList.java
```diff
@@ -0,0 +1,83 @@
...
+        map.computeIfAbsent(time, k -> new ArrayList<>()).addAll(events);
+    }
+
+    public void completed(Event e) {
+        completedMap.computeIfAbsent(e.time(), k -> new ArrayList<>()).add(e);
+    }
+
+
+    //read
+    public List<Event> getForTime(int time) {
```
vinopm: This can be package-private to avoid the possibility of code outside of the event package being able to modify the list
```suggestion
   List<Event> getForTime(int time) {
```

src/main/java/uk/co/ramp/event/EventList.java
```diff
@@ -0,0 +1,83 @@
...
+        try (Writer writer = new FileWriter("events.csv")) {
+            new CsvWriter().write(writer, finalList, ImmutableFormattedEvent.class);
+
+        } catch (IOException e) {
+            String message = "An error occurred while writing a CSV file";
+            LOGGER.error(message);
+            throw new CsvException(message, e);
+        }
+
+    }
```
vinopm: This could be moved to an EventOutputWriter.java to separate responsibilities

src/main/java/uk/co/ramp/event/EventList.java
```diff
@@ -0,0 +1,83 @@
...
+        return map;
+    }
+
+    public void output() {
+
+        List<Integer> timeStamps = new ArrayList<>(completedMap.keySet());
+        timeStamps.sort(Comparator.naturalOrder());
+        List<ImmutableFormattedEvent> finalList = timeStamps.stream().map(completedMap::get).flatMap(Collection::stream).map(formattedEventFactory::create).filter(Objects::nonNull).collect(Collectors.toList());
+
+        try (Writer writer = new FileWriter("events.csv")) {
```
vinopm: Pass in 'event.csv' file name via config or declare it as a static class variable

src/main/java/uk/co/ramp/event/EventProcessor.java
```diff
@@ -0,0 +1,378 @@
...
+        this.utils = utils;
+        this.eventList = eventList;
+        this.diseaseProperties = diseaseProperties;
+        this.isolationPolicy = isolationPolicy;
+    }
+
+
+    public void setPopulation(Map<Integer, Case> population) {
+        this.population = population;
+    }
```
vinopm, line 54: Would it be possible to pass the population object in through the constructor instead?

src/main/java/uk/co/ramp/event/EventProcessor.java
```diff
@@ -0,0 +1,378 @@
...
+
+        }
+        return Optional.empty();
+    }
+
+    private AlertStatus determineNextAlertStatus(AlertEvent event) {
+
+        Case thisCase = population.get(event.id());
+
+        List<AlertStatus> newStatus = getValidTransitions(event.newStatus());
```
vinopm: Perhaps calling newStatus as nextStatus would be more readable

src/main/java/uk/co/ramp/event/EventProcessor.java
```diff
@@ -0,0 +1,378 @@
...
+        switch (event.newStatus()) {
+            case AWAITING_RESULT:
+                return thisCase.isInfectious() ? TESTED_POSITIVE : TESTED_NEGATIVE;
+            case NONE:
+                return NONE;
+            case TESTED_POSITIVE:
+                return TESTED_POSITIVE;
+            default:
+                LOGGER.error(event);
+                throw new EventException("There is no case for the event" + event);
```
vinopm: Might be more clear if we call newStatus as nextStatus instead.

If the event.newStatus() is NONE, the nextstatus (based on valid transitions) are a list of [ALERTED, REQUESTED_TEST], but here we will always return NONE? (line 260)

vinopm: Also I think the case TESTED_POSITIVE is not reachable because of line 253

src/main/java/uk/co/ramp/event/EventProcessor.java
```diff
@@ -0,0 +1,378 @@
...
+            AlertStatus nextStatus = determineNextAlertStatus(event);
+
+            if (nextStatus != event.newStatus()) {
+                int nextTime = timeInStatus(nextStatus);
+
+                AlertEvent subsequentEvent = ImmutableAlertEvent.builder().
+                        id(event.id()).
+                        oldStatus(event.newStatus()).
+                        newStatus(nextStatus).
+                        time(time + nextTime).
```
vinopm: confusing to have it called nextTime, may better to call it nextTimeDelta?

src/main/java/uk/co/ramp/event/EventProcessor.java
vinopm: Also this skips the validity checking we have in place to check if the next transition is valid

src/main/java/uk/co/ramp/event/EventProcessor.java
```diff
@@ -0,0 +1,378 @@
...
+        if (newStatus.size() == 1) return newStatus.get(0);
+
+        // these cases have multiple paths
+        switch (event.newStatus()) {
+            case EXPOSED:
+                return determineInfection(population.get(event.id()));
+            case SYMPTOMATIC:
+                return determineSeverity(population.get(event.id()));
+            case SEVERELY_SYMPTOMATIC:
+                return determineOutcome(population.get(event.id()));
```
vinopm: This skips the validity checking that we have in place to check if the next transition is valid

src/main/java/uk/co/ramp/event/EventProcessor.java
```diff
@@ -0,0 +1,378 @@
...
+        // Look at all virus events
+        for (VirusEvent event : eventList.getForTime(time).stream().filter(event -> event instanceof VirusEvent).map(event -> (VirusEvent) event).collect(Collectors.toList())) {
+
+            population.get(event.id()).processEvent(event);
+            VirusStatus nextStatus = determineNextStatus(event);
+
+            // will return self if at DEAD or RECOVERED
+            if (event.newStatus() != nextStatus) {
+
+                int nextTime = timeInCompartment(event.newStatus(), nextStatus);
```
vinopm:
```suggestion
                int nextTimeDelta = timeInCompartment(event.newStatus(), nextStatus);
```

src/main/java/uk/co/ramp/event/EventProcessor.java
```diff
@@ -0,0 +1,378 @@
...
+
+                VirusEvent subsequentEvent = ImmutableVirusEvent.builder().
+                        id(event.id()).
+                        oldStatus(event.newStatus()).
+                        newStatus(nextStatus).
+                        time(time + nextTime).
```
vinopm:
```suggestion
                        time(time + nextTimeDelta).
```

src/main/java/uk/co/ramp/event/FormattedEventFactory.java
```diff
@@ -0,0 +1,101 @@
...
+import uk.co.ramp.people.Case;
+
+@Service
+public class FormattedEventFactory {
+
+    private static final Logger LOGGER = LogManager.getLogger(FormattedEventFactory.class);
+
+    public ImmutableFormattedEvent create(Event event) {
+
+        if (event instanceof VirusEvent) {
```
vinopm, line 16: The use of instanceof here seems like code smell. We may have to refactor but probably best to leave that for a separate PR as I think its not a simple fix

src/main/java/uk/co/ramp/event/FormattedEventFactory.java
```diff
@@ -0,0 +1,101 @@
...
+                eventType("AlertEvent").
+                id(event.id()).
+                newStatus(event.newStatus().toString()).
+                additionalInfo("").
+                build();
+    }
+
+
+    ImmutableFormattedEvent create(PolicyEvent event) {
+        // TODO: check when merging
```
vinopm: Should be fine to leave this as it is for now

src/test/java/uk/co/ramp/event/FormattedEventFactoryTest.java
```diff
@@ -0,0 +1,192 @@
...
+        Assert.assertNotNull(event.eventType());
+        Assert.assertEquals("PolicyEvent", event.eventType());
+
+        try {
+            formattedEventFactory.create(notCovered);
+
+        } catch (EventException e) {
+            Assert.assertThat(logSpy.getOutput(), containsString(notCovered.getClass().getSimpleName()));
+            throw e;
+        }
```
vinopm:
```suggestion
        assertThatExceptionOfType(EventException.class)
                .isThrownBy(() -> formattedEventFactory.create(notCovered))
                .withMessageContaining("Unknown Event Type")
                .withMessageContaining(notCovered.getClass().getSimpleName());
```

src/main/java/uk/co/ramp/Outbreak.java
```diff
@@ -63,171 +70,71 @@ public void setContactRecords(Map<Integer, List<ContactRecord>> contactRecords)
...
-
-    int calculateDailyStatistics(int time) {
+    void updateLogActiveCases(int time) {
+        int previousActiveCases = activeCases;
         Map<VirusStatus, Integer> stats = utils.getCmptCounts(population);
-        int activeCases = stats.get(EXPOSED) + stats.get(EXPOSED_2) + stats.get(INFECTED) + stats.get(INFECTED_SYMP);
+        activeCases = stats.get(EXPOSED) + stats.get(ASYMPTOMATIC) + stats.get(PRESYMPTOMATIC) + stats.get(SYMPTOMATIC) + stats.get(SEVERELY_SYMPTOMATIC);
 
-        CmptRecord cmptRecord = outputLog.log(time, stats);
+        CmptRecord cmptRecord = outputLog.log(time, stats, activeCases - previousActiveCases);
```
vinopm: I would push storing the state of activeCases into the implementation of LogDailyOutput. This would mean we can simply pass in (time, stats) as we were doing before, and let the LogDailyOutput class keep track of the previous cases.

src/main/java/uk/co/ramp/Outbreak.java
```diff
@@ -63,171 +70,71 @@ public void setContactRecords(Map<Integer, List<ContactRecord>> contactRecords)
...
-                    compartmentCounts.get(INFECTED) == 0 &&
-                    compartmentCounts.get(INFECTED_SYMP) == 0;
-
-
-            if (exitCondition) {
-                LOGGER.info("Steady state solution reached at t={} ", time);
-                LOGGER.info("Exiting early.");
-                return;
+            // stop random infections after contacts end
+            if (activeCases == 0 && lastContact < time) {
```
vinopm: I would calculate this on the fly instead of storing state in class.

Perhaps a separate private method to call from here:
```
    private int activeCases() {
        var stats = utils.getCmptCounts(population);
        return stats.get(EXPOSED) + stats.get(ASYMPTOMATIC) + stats.get(PRESYMPTOMATIC) + stats.get(SYMPTOMATIC) + stats.get(SEVERELY_SYMPTOMATIC);
    }
```

src/main/java/uk/co/ramp/Outbreak.java
```diff
@@ -63,171 +70,71 @@ public void setContactRecords(Map<Integer, List<ContactRecord>> contactRecords)
...
+                exposedBy(Case.getInitial()).
+                oldStatus(SUSCEPTIBLE).
+                newStatus(EXPOSED).
+                exposedTime(0).
+                id(-1).
+                time(0).build();
 
-            EvaluateCase evaluateCase = new EvaluateCase(population.get(id), diseaseProperties, distributionSampler);
-            evaluateCase.updateVirusStatus(EXPOSED, 0, Case.getInitial());
+        infectedIds.forEach(id -> virusEvents.add(ImmutableInfectionEvent.builder().from(genericEvent).id(id).build()));
```
vinopm: Personal preference but feel free to ignore:
```suggestion
        infectedIds.forEach(id -> virusEvents.add(ImmutableInfectionEvent.copyOf(genericEvent).withId(id)));
```

vinopm (commented):
@sam-brett I have reviewed all changes so far and left some suggestions/comments. Looks good to me on the whole, nice work with the event list. There are some uses of `instanceof` that I'd like us to avoid using, but perhaps we can address that in a follow-up PR. I'll review this PR again once the rest of your tests are in (as you mentioned previously).

src/main/java/uk/co/ramp/Outbreak.java
sam-brett: Fair point

src/main/java/uk/co/ramp/Outbreak.java
sam-brett: Yep that makes sense, although it does mean we have active cases being calculated in two places, but I don't think that matters.

src/main/java/uk/co/ramp/event/EventList.java
sam-brett: good call

src/main/java/uk/co/ramp/event/EventList.java
sam-brett: I've made it static for now, but it might be worth making an output config at some point so the output files are defined externally should the user want to pipe them elsewhere.

src/main/java/uk/co/ramp/Outbreak.java
vinopm: Nice. Yes you're right. Seems like a premature optimisation to me to calculate once and pass value through. My thinking was that this suggestion would simplify the code from a readability perspective

src/main/java/uk/co/ramp/event/EventProcessor.java
sam-brett: I guess we could, I think we'd need to make the population a bean first though.

src/main/java/uk/co/ramp/event/EventProcessor.java
sam-brett: changed to deltaTime

src/main/java/uk/co/ramp/event/EventProcessor.java
sam-brett: done

src/main/java/uk/co/ramp/event/EventProcessor.java
sam-brett: Added logic to the three methods

src/main/java/uk/co/ramp/event/FormattedEventFactory.java
sam-brett: Yeah I winced when I was writing it.

src/main/java/uk/co/ramp/people/Case.java
sam-brett: moved

src/main/java/uk/co/ramp/people/AlertStatus.java
sam-brett: oops. I added the validity check back into even processor.

sam-brett:
@vinopm I think that covers everything except the instanceOf calls.

src/test/java/uk/co/ramp/io/InfectionMapTest.java
```diff
@@ -1,14 +1,152 @@
...
+        infectionMap = mock(InfectionMap.class);
+        doCallRealMethod().when(infectionMap).outputMap(any());
+        doThrow(new IOException("")).when(infectionMap).recurseSet(anyList(), anyMap(), any(), anyInt());
+
+        try {
+            infectionMap.outputMap(new StringWriter());
+        } catch (InfectionMapException e) {
+            Assert.assertThat(logSpy.getOutput(), containsString("An error occurred while writing the map file"));
+            throw e;
+        }
```
vinopm:
```suggestion
assertThatExceptionOfType(InfectionMapException.class)
                .isThrownBy(() -> infectionMap.outputMap(new StringWriter()))
                .withMessageContaining("An error occurred while writing the map file");
```

src/test/java/uk/co/ramp/io/InfectionMapTest.java
```diff
@@ -1,14 +1,152 @@
...
+                stringWriter.toString().split("\n")).
+                filter(s -> !s.isBlank()).
+                map(s -> s.substring(0, s.indexOf("("))).
+                map(Integer::parseInt).
+                collect(Collectors.toSet());
+
+        Assert.assertEquals(lines, seeds);
+    }
+
+    @Test(expected = InfectionMapException.class)
```
vinopm:
```suggestion
```

Not needed if you use `assertThatExceptionOfType()`

src/main/java/uk/co/ramp/io/InfectionMap.java
```diff
@@ -89,5 +72,29 @@ void recurseSet(List<Case> target, Map<Integer, Set<Case>> infectors, Writer wri
 
     }
 
+    Map<Integer, Set<Case>> collectInfectors() {
+
+        List<Case> infections = population.values().stream()
+                .filter(c -> c.virusStatus() != SUSCEPTIBLE)
+                .sorted(Comparator.comparingInt(Case::exposedTime))
```
vinopm, line 79: I couldn't work out what the purpose of sorting this here is. We seem to be returning a set rather than a list?

src/main/java/uk/co/ramp/event/EventList.java
```diff
@@ -0,0 +1,89 @@
...
+
+
+    //read
+    List<Event> getForTime(int time) {
+        return map.getOrDefault(time, new ArrayList<>());
+    }
+
+    public Map<Integer, List<Event>> getMap() {
+        return map;
+    }
```
vinopm: This can be removed now:
```suggestion

```

vinopm (commented):
@sam-brett Thanks. I've marked the ones you have addressed as resolved. There are still 8 or so open comments/suggestions excluding the instanceof comment. Let me know what your thoughts are on it

vinopm:
@sam-brett There are still 8 or so open comments/suggestions excluding the instanceof comment. Let me know what your thoughts are on it

src/main/java/uk/co/ramp/event/EventProcessor.java
sam-brett: I'll remove that condition, as you're right about it being returned earlier. The transition is checked at line 116.

src/main/java/uk/co/ramp/io/InfectionMap.java
sam-brett: Replaced with list. This was to make sure the infections printed in time order, but obviously the set doesn't keep that so good spot.

src/main/java/uk/co/ramp/people/Case.java
sam-brett: Great spot. changed.

vinopm (approved):
Approved

PR #19, opened 2020-06-14
Bug fix for isolation policy

vinopm:
Previously if isolation for a policy e.g. 'stay at home policy' was set to 90% chance of isolation, instead of the probability of isolation being 0.9, we were isolating 100% of the time.
This is because the the threshold was set to a fixed number: '50' before. (50 is always less than 90). Now, instead, we compare the policy setting of 90% with a sampled value from a linear distribution between 0-100. This will ensure that we isolate 90% of cases going forward.

### DIFF
--- a/input/isolationPolicies.json
+++ b/input/isolationPolicies.json
@@ -129,5 +129,9 @@
     },
     "priority": 0
   },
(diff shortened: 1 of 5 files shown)


sam-brett (approved):
Simple changes, approved

PR #21, opened 2020-06-16
Fix infection empty map

vinopm:


sibylleM:
Works like a charm, thanks @vinopm!

sibylleM (approved):
Works like a charm! Thanks Vino

vinopm:
@sam-brett this is not needed anymore, as the other PR was merged which includes this change

PR #22, opened 2020-06-17
Specific isolation policies

vinopm:
Created policy configurations:

isolationPolicies.json is a symlink to isolationPolicies_Scenario1.json for now. It can be changed depending on the policy you wish to run the model with.

Scenario 0:
- No lockdown/isolation

Scenario 1:
- No lockdown (if proportion of population/R reaches certain value, we are not asking people to isolate)
- Isolation Policy: All Symptomatic & Severely symptomatic people are asked to isolate. All alerted contacts (people who have come into contact with someone infectious) are asked to isolate for a blanket 14 days (without caring about testing/targeting). The 14 days start as soon as the person starts to isolate.

Scenario 2:
- Lockdown policy: if over 10% of population are infected, we ask 90% of population to self-isolate regardless of symptoms. If 5-10% of population are infected, we ask 70% of population to self-isolate regardless of symptoms.
- Isolation Policy: Symptomatic people are asked to self-isolate for 13-14 days. The isolation time starts as soon as the person starts to isolate. People who are alerted that they've been in contact with someone infectious are asked to isolate for 13-14 days. Exception to this rule is if they test negative, then they do not need to continue isolating if the individual has not yet completed the 13-14 days for which they were asked to isolate.

Scenario 0 and 1 are what were discussed as first steps. Scenario 2 just exists to illustrate the current flexibility of the isolation policy framework that we have in place.

Please feel free to ask questions if this is not clear


sam-brett (commented):
Looks good from me. The only comment I would make is whether the symlink file should exist or if the input should be specified in inputLocations.json. This is purely styles, so I'm happy to approve as is providing @sibylleM agrees the polices match Louise's document.

sibylleM (approved):
Hi @vinopm looks also good to me. I can confirm scenario 0 and 1 matches what we discussed at the last meeting. Concerning scenario 2 I’ll have a think on how best to exploit this and will get back to you to discuss. Thanks a lot!

PR #23, opened 2020-06-17
Improve robustness to make sure input parameters are valid

vinopm:
Improve robustness to make sure input parameters are valid. This will be useful for when Rikiya/Fedor carry out inference work on our model. This will make sure inappropriate parameter values are not explored.


sam-brett (approved):
Approved. I've been using this library at work now as well, I like it a lot.

PR #24, opened 2020-06-19
Splitting up EventList (refactor PR - no functional changes)

vinopm:
One more refactor PR: Splitting up Event List into 4 event lists. One for each event type. 

No functional changes. 

Had to do this because we need to be able to extract all recent contact events for a given person when performing contact tracing. Therefore, we need to be able to extract the contact events form the event list. So I've split the logic so that there is an event list per event type. 

I will raise my functional changes PR for contact tracing once this is reviewed/merged.

Thanks


sam-brett (approved):
Lots of churn, but it all looks good to me.

vinopm:
Thanks @sam-brett, I appreciate the review

PR #26, opened 2020-06-23
Rikiya exposure model reparametrise

maxlogutility:
Proposing to change the exposure probability from contact weight divided by `exposureTuning` into the proposed one involving two parameters `exposureProbability4UnitContact` and `exposureExponent`. The rationale and detail of the new parametrisation are provided in Javadoc of ContactEventProcessor#evaluateExposures.
This update should exclude the weird behaviour such that contact weight divided by exposureTuning can exceed 1, and also enables tuning of sensitivity of the probability to contact weight, through the new parameter `exposureExponent`.

I also thought of introducing randomness in determining weak contact by using `exposureThreshold`. This rule, however, seems to be intended as a deterministic policy, and hence I kept it unchanged.


vinopm:
@maxlogutility Hi Rikiya, please can you resolve the conflicts and also remove the addition of application.log files and infectionMap.txt from the PR. Thanks

bin/.gitignore
```diff
@@ -0,0 +1,2 @@
+/main/
+/test/
```
vinopm: I think `bin/` should be in the root .gitignore file instead

src/main/java/uk/co/ramp/event/ContactEventProcessor.java
```diff
@@ -66,27 +67,52 @@ private boolean isContactIsolated(ContactEvent contact, Case caseA, Case caseB,
...
+	 * @return
+	 */
+	Optional<InfectionEvent> evaluateExposures(ContactEvent c, int time) {
+		final Case personA = getMostSevere(population.get(c.to()), population.get(c.from()));
+		final Case personB = personA == population.get(c.to()) ? population.get(c.from()) : population.get(c.to());
+
+		final boolean dangerMix = personA.isInfectious() && personB.virusStatus() == SUSCEPTIBLE;
+		
+		final double expBias = diseaseProperties.exposureProbability4UnitContact()/(1.0-diseaseProperties.exposureProbability4UnitContact());
+		final double exposureProb = 1. / (1. + 1./(expBias*FastMath.pow(c.weight(), diseaseProperties.exposureExponent())));
```
vinopm, line 103: Probably best if @sibylleM can review the math here

src/main/java/uk/co/ramp/event/ContactEventProcessor.java
```diff
@@ -66,27 +67,52 @@ private boolean isContactIsolated(ContactEvent contact, Case caseA, Case caseB,
...
+
+		if (dangerMix&& distributionSampler.uniformBetweenZeroAndOne() < exposureProb) {
+			LOGGER.debug("       DANGER MIX");
```
vinopm: Probably best if @sibylleM can review the math here

src/main/java/uk/co/ramp/event/ContactEventProcessor.java
```diff
@@ -66,27 +67,52 @@ private boolean isContactIsolated(ContactEvent contact, Case caseA, Case caseB,
...
+	 * 
+	 * @param c
+	 * @param time
+	 * @return
+	 */
+	Optional<InfectionEvent> evaluateExposures(ContactEvent c, int time) {
+		final Case personA = getMostSevere(population.get(c.to()), population.get(c.from()));
+		final Case personB = personA == population.get(c.to()) ? population.get(c.from()) : population.get(c.to());
+
+		final boolean dangerMix = personA.isInfectious() && personB.virusStatus() == SUSCEPTIBLE;
```
vinopm: `final`'s are not needed here as local variables within methods are 'effectively final' here. http://ilkinulas.github.io/programming/java/2016/03/27/effectively-final-java.html
https://stackoverflow.com/questions/20938095/difference-between-final-and-effectively-final

bin/.gitignore
maxlogutility: Just added `bin` in top `.gitignore` and removed `bin/.gitignore`.

src/main/java/uk/co/ramp/event/ContactEventProcessor.java
maxlogutility: Thanks for the information. I just applied my old two cents that old JVMs were working slightly faster when I place `final` if the variable is guaranteed to unchange. Also it was my personal practice to avoid to accidentally substitute a new value into it.

Just removed the `final` but retaining it will also be not harmful.

maxlogutility:
Now passed all checks, with loosening the criteria in exponential decay approximation.
Probability of exposure is in a similar level to the previous model using contact_weight / 100.

sibylleM (approved):
Hi @maxlogutility and @vinop, the equations look sound to me. I think it definitely makes more sense to use this than what we had before.

src/main/java/uk/co/ramp/event/ContactEventProcessor.java
sibylleM: @vinopm @maxlogutility Looks good to me.

infectionMap.txt
```diff
@@ -0,0 +1,32 @@
+31(0)         ->  [55(6)]
```
vinopm: This file should be removed :)

src/test/java/uk/co/ramp/OutbreakTest.java
```diff
@@ -97,9 +97,17 @@ public void runContactDataWithRandoms() {
...
+    //
+    // If r='randomInfection' ratio of people become other status than SUSCEPTIBLE,
+    // after T days, remaining population of SUSCEPTIBLE persons should be
+    // (1-r)^T = exp(T log(1-r))
+    // Previously, an approximation log(1-r)=r was applied but this is valid only when r is small
+    double test = popSize * Math.exp(Math.log(1. - randomInfection) * days);
+
+    // Large tolerance 0.5 should be allowed given the mechanism of contact-based exposure, in
+    // addition to the random infection
+    Assert.assertEquals(test / (double) popSize, sus / (double) popSize, 0.5);
```
vinopm: If we run this for larger sample size, can we reduce the tolerance? 0.5 is very large. Always better to have a narrow tolerance in tests so it is easy to detect when something breaks. What are your thoughts on this? Happy to discuss further.

src/main/java/uk/co/ramp/event/ContactEventProcessor.java
```diff
@@ -76,16 +77,47 @@ private boolean isContactIsolated(
...
+   * == 0.
+   *
+   * <p>The bias and exponent parameters control level of the probability and sensitivity of the
+   * probability to the contact weight, respectively.
+   *
+   * <p>For an intuitive parameter specification in {@link DiseaseProperties}, instead of specifying
+   * bias in [-infty, infty], probability U == P(exposure=1|C=1) = {@link
+   * exposureProbability4UnitContact#DiseaseProperties}, whose support is (0, 1), should be
+   * provided. Then it is automatically converted into exp(bias) = U/(1-U) based on Sigmoid(bias) =
+   * U.
```
vinopm, line 101: 👍 thanks for the explanation

src/test/java/uk/co/ramp/OutbreakTest.java
maxlogutility: I do not think larger sample size (do you mean population?) ends in lower discrepancy. It is essentially from the model structure in infection, not from randomness.

0.5 is indeed large but under the new model 0.1 cannot be achieved because there are new infections by both random without-contact one and contact-based person-to-person one. When the impact of the latter gets stronger, exponential decay approximation is very erroneous.

One alternative considerable is to calibrate the logarithmic decay rate and set the tolerance on
the difference between `-Math.log(test / popSize) / days` and `-Math.log(1-randomInfection)`. Another alternative is to calibrate the difference in early period. If early-stage exposure involves less second-order effect than the late-stage one, so exponential decay should have a higher approximation accuracy. Instead of taking the 11 days, we can try to pick up the first 3 or 4 days and calibrate the decay rate in logarithmic sense. Detection of crucially wrong modification should be possible even in this simplified version.

vinopm (approved):
Nice, looks good to me

src/test/java/uk/co/ramp/OutbreakTest.java
```diff
@@ -97,54 +97,63 @@ public void runContactDataWithRandoms() {
...
+    // after T days, remaining population of SUSCEPTIBLE persons should be
+    // (1-r)^T = exp(T log(1-r))
+    // Previously, an approximation log(1-r)=r was applied but this is valid only
+    // when r is small
+    // double test = popSize * Math.exp(Math.log(1. - randomInfection) * days);
+
+    // Calibrate the error not in the population itself but in the logarithmic rate space.
+    Assert.assertEquals(
+        -Math.log(sus / (double) popSize) / days, -Math.log(1 - randomInfection), 0.1);
   }
```
maxlogutility: @vinopm This is one possible alternative. In the current test, approximately 0.12 and 0.1 are the values inside the first and second argements in `Asset.assertEquals`, respectively. We can use even smaller tolerance if using geometric rate. Another alternative is to use tolerance proportional to randomInfection, say, `0.5 * randomInfection`. However this is problematic because even if `randomInfection=0` virus diffuses purely by contact and number of `SUSCEPTIBLE` persons decline over time. So using a constant rather than proportional one is commendable.

vinopm: Thanks for the explanation RIkiya. So we could change the tolerance to say 0.05?

maxlogutility: Yes, if you are happy I will set it into 0.05. Also will remove commented-out old codes for simplification.

vinopm: Sure, I am happy with that

vinopm:
Hi @sam-brett , @johnnonweiler , if you're happy with this, could you please merge this in? Thanks!

PR #28, opened 2020-06-24
Bug fix for random infection generation and initial infection events

vinopm:
Previously they were being read in and ignored. Now they are processed.


vinopm:
@sam-brett had to fix a couple of tests, PR is ready now. Thanks

sam-brett (approved):

PR #29, opened 2020-06-24
Contact tracing implementation + multiple tracing levels + isolate from time of contact

vinopm:
Implementation of contact tracing by setting recent contacts (whether they were infected or not doesn't matter) to ALERTED and setting infector to REQUESTED_TEST


vinopm:
@sam-brett note: to be reviewed/merged after this PR https://github.com/ScottishCovidResponse/Contact-Tracing-Model/pull/28

sam-brett (commented):
I think I screwed up with the Event List, at least some of this functionality was there before. The code logic all looks good at a glance, although I'm wondering if we need to adjust the events output to account for this. Previously the AlertEvent printed out the suspected infector and contact time in the additional info column. I think that could be useful to include going forwards. 

My only general comment would be that some of the more complex streams could use a comment

vinopm:
Thanks @sam-brett. I'll add some comments for the more complex streams. 

Are you referring to the events.csv output file? I believe we are still printing out the suspected infector and infection time in the additionalInfo column. Have a look at `FormattedEventFactory.java`

Do you have any thoughts on if there's any other functionality which may have been accidentally removed?

sam-brett:
@Vino ignore me, I'm having a brain dead day. I was using excel and filtering the wrong thing. I've just followed a few cases through the event lists and they seem to make sense. I think this highlights that we maybe need some regression/higher level testing in a few places to catch these things.

vinopm:
@sam-brett No problem. That's great. Thanks for trying out a few cases. Yes completely agree with you on the higher level integration/regression testing. I'll add the comments as you suggested and then let you know. I'll probably have this done tomorrow morning

vinopm:
@sam-brett updated with comments as you suggested

vinopm:
@sam-brett added a bug fix where alert events are generated based off of current event that was just processed instead of the future event. 
Also added implementation which would allow for multiple levels of contact tracing.
Please could you review/merge. Thanks

vinopm:
@sam-brett Added some additional things into this PR
1. Added a bug fix. Now alert events are generated based off of the current event that was just processed instead of the future (subsequent) event. 
2. Added implementation which would allow for multiple levels of contact tracing.
3. Added ability to count isolation days from time of contact based on 'exposedTime' rather than alert time

This is ready for review again. Please could you review/merge. Thanks

src/main/java/uk/co/ramp/policy/alert/AlertPolicyContext.java
```diff
@@ -0,0 +1,44 @@
...
+import org.apache.logging.log4j.Logger;
+import org.springframework.boot.SpringBootConfiguration;
+import org.springframework.context.annotation.Bean;
+import uk.co.ramp.ConfigurationException;
+import uk.co.ramp.Population;
+import uk.co.ramp.event.CompletionEventListGroup;
+
+@SpringBootConfiguration
+public class AlertPolicyContext {
+  private static final String ALERT_SETTINGS_LOCATION = "input/alertPolicies.json";
```
sam-brett: Not sure if we should be keeping these policies hard coded now we have an inputs file? Unless you think this should be here because it is context specific? I think the same goes for the isolation inputs.

src/main/java/uk/co/ramp/policy/isolation/SingleCaseIsolationPolicy.java
```diff
@@ -142,8 +144,30 @@ private IsolationMapValue updatedMapValue(
...
 
+  private int startTime(
+      IsolationStartTimeType policyStartTimeType, int currentTime, int contactTime) {
+    if (policyStartTimeType == IsolationStartTimeType.ABSOLUTE) {
+      return currentTime;
+    } else if (policyStartTimeType == IsolationStartTimeType.CONTACT_TIME) {
+      return contactTime;
+    } else {
+      throw new IllegalStateException("Isolation Start Time type is invalid.");
+    }
```
sam-brett, line 156: Would a switch statement look cleaner here as there are only two cases from an Enum?

src/main/java/uk/co/ramp/event/EventContext.java
```diff
@@ -7,20 +7,32 @@
 import uk.co.ramp.event.types.*;
 import uk.co.ramp.io.InitialCaseReader;
 import uk.co.ramp.io.types.DiseaseProperties;
-import uk.co.ramp.policy.IsolationPolicy;
+import uk.co.ramp.policy.alert.AlertChecker;
+import uk.co.ramp.policy.isolation.IsolationPolicy;
 
 @SpringBootConfiguration
 public class EventContext {
   private final EventList<Event> completedEventList = new EventList<>();
```
sam-brett: Unused

src/main/java/uk/co/ramp/event/types/CommonVirusEvent.java
```diff
@@ -8,4 +11,20 @@
   VirusStatus oldStatus();
 
   VirusStatus nextStatus();
+
+  @Value.Check
+  default void check() {
+    Preconditions.checkState(
+        checkValidTransition(),
+        "Transition from " + oldStatus() + " to " + nextStatus() + " is not valid");
```
sam-brett:
```suggestion
      "Transition from %s to %s is not valid", oldStatus(), nextStatus());
```

I think this is more efficient.

sam-brett (changes requested):
There are some tiny suggestions, but you may disagree with them. Otherwise it looks great.

vinopm:
Thanks Sam, all seems like very sensible suggestions to me. Will update and then let you know. Thanks!

src/main/java/uk/co/ramp/event/EventContext.java
vinopm: Good spot, thanks!

src/main/java/uk/co/ramp/policy/alert/AlertPolicyContext.java
vinopm: Yep agreed, both should move over to the inputs file. I'll update this

vinopm:
@sam-brett updated with your suggestions now.

sam-brett (approved):

PR #31, opened 2020-06-30
Accounting for floating point error on the population distribution.

maxlogutility:
For some random population distribution values exact matching of sum to be one leads termination of the run, due to the floating point approximation error. Here a small tolerance is introduced for preventing the termination during the repeated drawing of random population distributions in sensitivity analysis.

### DIFF
--- a/src/main/java/uk/co/ramp/io/types/PopulationProperties.java
+++ b/src/main/java/uk/co/ramp/io/types/PopulationProperties.java
@@ -19,8 +19,9 @@ public interface PopulationProperties {
 
   @Check
   default void check() {
+    final double eps = 1e-6;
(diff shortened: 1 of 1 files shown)


vinopm (approved):
LGTM

src/main/java/uk/co/ramp/io/types/PopulationProperties.java
```diff
@@ -19,8 +19,9 @@
 
   @Check
   default void check() {
+    final double eps = 1e-6;
```
vinopm, line 22: Final not needed here, but it's ok, no harm.

johnnonweiler (approved):

PR #33, opened 2020-07-01
Changes to reflect feedback from last model meeting on Tuesday 30 June

vinopm:
These changes have been produced on the back off the feedback from Louise and @sibylleM in the model meeting on Tuesday.

Changes in this PR outlined below:

1. Symlink removed so windows filesystems are supported now.
2. New command line argument specifiable: --overrideInputFolderLocation=="input/scenarios/scenario0" to test scenario 0.
3. By default an internal random seed is used now. It can be overridden to a fixed seed in either the runSettings.json or via command line: --seed=5
4. Renamed AlertPolicy to TracingPolicy so that it is more understandable to reader
5. Introduced extra configuration. Previously only symptomatic people whose alert status was NONE were going through a cycle of contact tracing. Now it is configurable to allow for other scenarios such as invoking the contact tracing when someone is tested positive rather than when they are symptomatic
6. Using id csv data file now to load ages. If id of contact is not found in this file, we resort back to the specified distribution (as before).
7. Tracing 'delay' parameter introduced.


sibylleM (approved):
Thanks @vinopm for adding the changes so speedily and for adding the readme files for each policy scenario. This is makes it so much easier to keep track which scenario we are running.

vinopm:
@johnnonweiler @sam-brett could I have your review please? Thank you

johnnonweiler (approved):
This looks good to me.  I have checked that it compiles on Windows.  I have not thoroughly reviewed most of the changes, and they could definitely do with some review from @sam-brett when he's back.

The test `testInvalidFile()` failed on my machine, presumably due to me using the wrong version of Java - I got "This file doesn't exist (The system cannot find the file specified)", where the test expected "This file doesn't exist (No such file or directory)".

Would it be possible to make a custom deserializer for the Distribution class?  Then you could improve the way that "FLAT" distributions are represented in the JSON.

src/main/java/uk/co/ramp/AppConfig.java
```diff
@@ -80,35 +105,53 @@ public PopulationProperties populationProperties() {
...
-      RandomDataGenerator r = new RandomDataGenerator();
-      r.reSeed(standardProperties().seed() + arg);
-      return r;
+      OptionalInt cmdLineSeedOverride =
+          Stream.ofNullable(seed).mapToInt(Integer::parseInt).findAny();
+      OptionalInt propertiesSeedOverride = standardProperties().seed();
+      Stream.of(cmdLineSeedOverride, propertiesSeedOverride)
+          .flatMapToInt(OptionalInt::stream)
+          .reduce(Integer::sum)
+          .ifPresentOrElse(seed -> useSeed(rdg, seed), this::logUsingDefaultSeed);
```
johnnonweiler: This is quite different to the style I'm used to (but I'm not complaining)!
I would have done something like:
```
if(seed == null && !standardProperties().seed().isPresent()) {
    logUsingDefaultSeed();
} else {
    int cmdLineSeed = seed == null ? 0 : Integer.parseInt(seed);
    int propertiesSeed = standardProperties().seed().orElse(0);
    useSeed(rdg, cmdLineSeed + propertiesSeed);
}
```

vinopm: Seems like I've overused the notion of streams here, definitely simpler to read your version. Will update. Thank you!

vinopm:
@johnnonweiler requested for your review, made a few more changes. Will merge once you're happy with the new changes.
Thanks

src/main/java/uk/co/ramp/event/AlertEventProcessor.java
```diff
@@ -89,6 +97,9 @@ int timeInStatus(AlertStatus newStatus) {
 
     MeanMax progressionData;
     switch (newStatus) {
+      case TESTED_POSITIVE:
+        progressionData = ImmutableMeanMax.builder().mean(1).max(1).build();
+        break;
```
johnnonweiler, line 102: (At a glance,) this looks like it could be more concise:
```java
     case TESTED_POSITIVE:
     case TESTED_NEGATIVE:
     case ALERTED:
        progressionData = ImmutableMeanMax.builder().mean(1).max(1).build();
        break;
```

src/main/java/uk/co/ramp/policy/alert/AlertChecker.java
```diff
@@ -47,24 +56,36 @@
...
+    if (tracingPolicyItem.isEmpty()) {
+      return requestTestForReporterEvent;
+    }
+    var startTime = currentTime - tracingPolicyItem.get().recentContactsLookBackTime() + 1;
 
     var alertsFromRecentContacts =
         alertContactTracer.traceRecentContacts(startTime, currentTime, personId).stream()
-            .filter(a -> currentReporterAlertStatus == tracingPolicyReporterAlertStatus)
-            .filter(a -> nextVirusStatus == tracingPolicyReporterVirusStatus)
+            .filter(a -> shouldPerformTracingForThisLink(tracingPolicyItem.get()))
```
johnnonweiler, line 68: Similar question to the above.  I'm clearly missing something here!

johnnonweiler (approved):
The bits I understand look ok, but I'm afraid there's a lot I don't understand here, and I'm not going to have time today or tomorrow to figure it out.  I suggest you merge now, and then make improvements in a future PR following review from @sam-brett.

src/main/java/uk/co/ramp/AppConfig.java
johnnonweiler: Thanks!

src/main/java/uk/co/ramp/policy/alert/AlertChecker.java
johnnonweiler: I can't see my "above" comment.  The question is what is this filter doing?  I find it strange that `a` seems to be unused.

src/main/java/uk/co/ramp/event/AlertEventProcessor.java
vinopm: Thanks @johnnonweiler good spot, will update in a separate PR. Thanks

src/main/java/uk/co/ramp/policy/alert/AlertChecker.java
vinopm: Hi @johnnonweiler, it's carrying out filtering which is based on external variables rather than the recent contact itself. I will update this so that it is more clear in a follow up PR. Thanks

PR #36, opened 2020-07-07
Parameterizing for output folder #568

sam-brett:
Works in the same way as your input config. I've made the default an "output" folder, for consistency.


src/main/java/uk/co/ramp/AppConfig.java
```diff
@@ -33,16 +35,34 @@
...
-      @Value("${overrideInputFolderLocation:#{null}}") String overrideInputFolderLocation) {
+      @Value("${overrideInputFolderLocation:#{null}}") String overrideInputFolderLocation,
+      @Value("${overrideOutputFolderLocation:#{null}}") String overrideOutputFolderLocation) {
     this.seed = seed;
     this.overrideInputFolderLocation = overrideInputFolderLocation;
+    this.overrideOutputFolderLocation = overrideOutputFolderLocation;
+  }
+
+  @Bean
+  public File outputFolder() {
```
vinopm: I think with Beans, the scope is singleton, so you'd only be able to ever have one `File` instance which could be injected in multiple places. This would become problematic if we want to define another Bean with a return type of `File`. As `File` is a very generic class, I think it would be better to make the return type more specific. Perhaps a wrapper class called `OutputFiles` would do?

vinopm (commented):
Nice work. All looks good to me, left a suggestion/comment for you.

Thanks

vinopm (approved):

PR #37, opened 2020-07-07
New Compliance parameter

sam-brett:


sam-brett:
Hi @Vino not sure if this covers everything or I've missed a subtlety somewhere, but I think this is what the ticket asked for. 😬

src/main/java/uk/co/ramp/event/ContactEventProcessor.java
```diff
@@ -100,12 +117,12 @@ private boolean isContactIsolated(
...
    * U.
    *
+   * @param personA the most severe case
    * @param c
    * @param time
    * @return
    */
-  Optional<InfectionEvent> evaluateExposures(ContactEvent c, int time) {
-    Case personA = getMostSevere(population.get(c.to()), population.get(c.from()));
+  Optional<InfectionEvent> evaluateExposures(Case personA, ContactEvent c, int time) {
```
vinopm: It would be cleaner if we either pass in as args `Case victim, Case potentialInfector, int time` or `ContactEvent c, int time` instead of both personA and c, where c already contains information about personA.
For the latter option we could derive the victim/infector again using `getMostSevere`.

src/test/java/uk/co/ramp/event/ContactEventProcessorTest.java
```diff
@@ -93,6 +97,95 @@ public void evaluateContact() {
...
+    IsolationPolicy isolationPolicy = Mockito.mock(IsolationPolicy.class);
+    when(isolationPolicy.isContactIsolated(any(), any(), anyDouble(), anyDouble(), anyInt()))
+        .thenReturn(true);
+
+    ReflectionTestUtils.setField(eventProcessor, "isolationPolicy", isolationPolicy);
+
+    Case mock0 = mock(Case.class);
+    when(mock0.virusStatus()).thenReturn(SYMPTOMATIC);
+    when(mock0.isInfectious()).thenReturn(true);
+    when(mock0.isolationCompliance()).thenReturn(-0.5);
```
vinopm: Should the expected range of compliance values be between 0 and 1?

src/main/java/uk/co/ramp/event/ContactEventProcessor.java
```diff
@@ -59,19 +64,31 @@ private boolean isContactIsolated(
...
 
-    if (shouldIsolateContact) {
+    boolean compliesWithIsolation = true;
+    // check for whether infector would isolate
+    if (distributionSampler.uniformBetweenZeroAndOne() < potentialSpreader.isolationCompliance()) {
+      LOGGER.info("person with id %d is not complying with isolation policy for this contact.");
+      compliesWithIsolation = false;
+    }
+
+    if (shouldIsolateContact && compliesWithIsolation) {
```
vinopm: Isolation compliance has already been implemented. If you have a look at `SingleCaseIsolationPolicy.java` around line 185, you will see that the `isContactIsolated()` call in from this file on line 75 already captures this.

We were previously using the 'compliance', as opposed to 'isolationCompliance', but you have implemented the name change correctly for that in this PR. I think the check here should be removed though

vinopm (changes requested):
Thanks @sam-brett, Please see my comments and let me know what you think. I think I may have not been clear enough on the original ticket!

We still need to have another parameter called 'reportingCompliance' and this is to be inserted in `VirusEventProcessor.java` line 53. Probably should be a if conditional block around the `checkForAlert` call, where we only do the check if the person's reportingCompliance value is high enough.

src/main/java/uk/co/ramp/event/ContactEventProcessor.java
sam-brett: No problem. Will change.

src/main/java/uk/co/ramp/event/ContactEventProcessor.java
sam-brett: ah fair enough I'll remove that then.

src/test/java/uk/co/ramp/event/ContactEventProcessorTest.java
sam-brett: The -ve is just to ensure the conditional fails. It saved mocking the sampler, this will be deleted anyway.

sam-brett:
@vinopm That makes more sense. I think my brain is still spinning up a little bit. Hopefully this is more like what you meant. Thanks

src/main/java/uk/co/ramp/event/VirusEventProcessor.java
```diff
@@ -50,8 +61,14 @@ public ProcessedEventResult processEvent(VirusEvent event) {
               .time(event.time() + deltaTime)
               .build();
 
-      List<AlertEvent> alertEvents = checkForAlert(event.id(), event.time());
-
+      List<AlertEvent> alertEvents;
+      if (event.nextStatus() == VirusStatus.SYMPTOMATIC
+          && thisCase.reportingCompliance() > distributionSampler.uniformBetweenZeroAndOne()) {
```
vinopm: Hi @sam-brett, there is already a check for if the person is symptomatic within the `checkForAlert` call (in AlertChecker.java). So I think that part of the conditional statement can be removed, so that we ensure code isn't duplicated

```suggestion
      if (thisCase.reportingCompliance() > distributionSampler.uniformBetweenZeroAndOne()) {
```

vinopm (changes requested):

sam-brett:
Not sure why this is still showing as changes requested, but I've committed your suggestion.

vinopm (approved):

PR #39, opened 2020-07-10
Adding Time steps per day

sam-brett:
Initial implementation to allow multiple timesteps per day. 

Tests need updating


src/main/java/uk/co/ramp/Outbreak.java
```diff
@@ -42,7 +38,7 @@
   private final File outputFolder;
 
   private final Population population;
-  private final Map<Integer, CmptRecord> records = new HashMap<>();
+  private final Map<Double, CmptRecord> records = new HashMap<>();
```
vinopm: @sam-brett It's dangerous to use double for a key in a hashmap because of floating point precision issues. You will not always be able to look up a key as the comparison may be slightly off and will not always be equal. 

We should keep integer as the key type and then scale based on number of timesteps when we access or put anything into the map

vinopm (changes requested):

input/inputLocations.json
```diff
@@ -5,6 +5,6 @@
   "contactData": "homogeneous_contacts.csv",
   "ageData": "ids_Paul.csv",
   "initialExposures": "initialExposures.csv",
-  "tracingPolicies": "tracingPolicies.json",
-  "isolationPolicies": "isolationPolicies.json"
+  "tracingPolicies": "scenarios/scenario1a/tracingPolicies.json",
+  "isolationPolicies": "isolationPolicies_Example.json"
```
vinopm: These values shouldn't be changed in this way because the values are used to determine the filenames to look for when the input folder is overridden. 

If we want to use scenario1a's tracing Policy as the default one, and the example isolation policy file as the default one, i would recommend copying over (and replacing) the tracingPolicies.json file in scenario 1a to the default location: "tracingPolicies.json". Likewise with the isolation policy. Copy over isolationPolicies_Example.json -> isolationPolicies.json

src/main/java/uk/co/ramp/event/AlertEventProcessor.java
```diff
@@ -120,6 +117,6 @@ int timeInStatus(AlertStatus newStatus) {
             .mean(progressionData.mean())
             .max(progressionData.max())
             .build();
-    return distributionSampler.getDistributionValue(distribution);
+    return distributionSampler.getDistributionValue(distribution) * properties.timeStepsPerDay();
```
vinopm, line 120: What is the reason for scaling here as opposed to scaling up the distribution mean/max values in diseaseSettings.json? i.e. if timeStepsPerDay is 2, then we could double all values in the diseaseSettings.json distributions? 

This would allow for mean/max values to represent half days as well. If we scale here, there is no way to pass in half day time step via diseaseSettings.json

src/main/java/uk/co/ramp/io/ContactReader.java
```diff
@@ -15,21 +20,54 @@
...
-        .filter(event -> event.time() <= properties.timeLimit())
+        .filter(event -> event.time() <= properties.timeLimit() * properties.timeStepsPerDay())
         .collect(Collectors.toUnmodifiableList());
   }
+
+  List<ImmutableContactEvent> resampleContacts(List<ImmutableContactEvent> contactEvents) {
+    int steps = properties.timeStepsPerDay();
+
+    Set<Integer> timeSteps =
+        contactEvents.stream().map(ImmutableContactEvent::time).collect(Collectors.toSet());
```
vinopm, line 50: Should probably convert this to a sorted list after because we are iterating through it, and we would want to iterate through in time step order?

src/main/java/uk/co/ramp/io/ContactReader.java
```diff
@@ -15,21 +20,54 @@
...
+      List<ImmutableContactEvent> timeEvents =
+          contactEvents.stream().filter(event -> event.time() == step).collect(Collectors.toList());
+
+      int start = step * steps;
+      int end = start + steps - 1;
+      int[] outcomes = IntStream.rangeClosed(start, end).toArray();
+
+      EnumeratedIntegerDistribution distribution =
+          new EnumeratedIntegerDistribution(
+              dataGenerator.getRandomGenerator(), outcomes, properties.timeStepSpread());
```
vinopm: Would be better to move this functionality into DistributionSampler under a new method which would return the sample. Then pass in our DistributionSampler into this class instead of RandomDataGenerator

src/main/java/uk/co/ramp/io/ContactReader.java
```diff
@@ -15,21 +20,54 @@
...
     this.properties = standardProperties;
+    this.dataGenerator = dataGenerator;
   }
 
   public List<ContactEvent> readEvents(Reader reader) throws IOException {
 
     List<ImmutableContactEvent> contactEvents =
         new CsvReader().read(reader, ImmutableContactEvent.class);
 
+    if (properties.timeStepsPerDay() > 1) {
```
vinopm: Seems like this if condition is not actually necessary (premature optimisation)

src/main/java/uk/co/ramp/io/LogDailyOutput.java
```diff
@@ -16,7 +16,7 @@
   private static final Logger LOGGER = LogManager.getLogger(LogDailyOutput.class);
   private int previousActiveCases = 0;
 
-  public CmptRecord log(int time, Map<VirusStatus, Integer> stats) {
+  public CmptRecord log(int time, Map<VirusStatus, Integer> stats, int timeStepsPerDay) {
```
vinopm: Pass in Properties through constructor and store the properties.timeStepsPerDay as private final instance variable. The value for this is constant for a given run of the model, so it doesn't need to be passed in as a parameter into the log method

src/test/java/uk/co/ramp/event/AlertEventProcessorTest.java
```diff
@@ -33,6 +32,7 @@
 
   private DiseaseProperties diseaseProperties;
   private Population population;
+  @Autowired private StandardProperties properties;
```
vinopm, line 35: We should try to avoid use of Autowired in unit tests. It would be better to mock StandardProperties + provide when() + thenReturn() definitions via mockito. This way we can make sure that the functionality of AlertEventProcessor is tested as an individual unit, instead of the test including running code outside the scope of the AlertEventProcessor class. 

I believe for this case we just need to remove:
```
  private StandardProperties properties;
```
and add into setUp

```
    properties = mock(StandardProperties.class);
    when(properties.timeStepsPerDay()).thenReturn(2);
```

src/test/java/uk/co/ramp/event/CommonVirusEventProcessorTest.java
```diff
@@ -39,13 +42,13 @@
   private CommonVirusEventProcessor<Event> eventProcessor;
 
   private DiseaseProperties diseaseProperties;
+  @Autowired private StandardProperties properties;
```
vinopm, line 45: Please see my other comment about preferring mocked objects to use of Autowired in unit tests.

src/test/java/uk/co/ramp/event/InfectionEventProcessorTest.java
```diff
@@ -43,6 +44,7 @@
   private InfectionEventProcessor eventProcessor;
   private DiseaseProperties diseaseProperties;
 
+  @Autowired private StandardProperties properties;
```
vinopm, line 48: Please see my other comment about preferring mocked objects to use of Autowired in unit tests.

src/test/java/uk/co/ramp/io/readers/StandardPropertiesReaderTest.java
```diff
@@ -15,7 +15,9 @@
           + "  'timeLimit': 100,"
           + "  'initialExposures': 1000,"
           + "  'seed': 0,"
-          + "  'steadyState': true"
+          + "  'steadyState': true,"
+          + "  'timeStepsPerDay': 4,"
+          + "  'timeStepSpread': [0.25,0.25,0.25,0.25]"
```
vinopm, line 20: Missing assertions for these in testRead()

src/test/java/uk/co/ramp/AppConfigTest.java
```diff
@@ -161,7 +161,9 @@ public void randomDataGeneratorNoArgs() throws ConfigurationException {
         containsString("Additional Seed information not provided, using internal random seed."));
   }
 
+  // TODO: can't see why this is failing... it works in debug for me.
```
vinopm: If you add 
  "seed": 123,
within the testSeedOverride/runSettings.json file, it should fix the test

vinopm (changes requested):

src/main/java/uk/co/ramp/event/AlertEventProcessor.java
sam-brett: I thought the idea was that we kept the disease settings in days, then scaled everything else. I think it might be cleaner to change the MeanMax interface to be doubles rather than ints so we can add finer granularity there.

vinopm: Sure that sounds right. Sounds good to me.

src/test/java/uk/co/ramp/event/AlertEventProcessorTest.java
sam-brett: There are a few of these to remove. I've got a tech debt branch so I'll include these on there.

src/main/java/uk/co/ramp/distribution/DistributionSampler.java
```diff
@@ -48,4 +49,9 @@ public int getDistributionValue(Distribution distribution) {
 
     return value;
   }
+
+  public EnumeratedIntegerDistribution resampleDays(int[] outcomes, double[] timeSpread) {
+
+    return new EnumeratedIntegerDistribution(rng.getRandomGenerator(), outcomes, timeSpread);
```
vinopm: Could we return a sampled value here instead of returning the distribution?

sam-brett: wouldn't that mean we would be creating a new distribution for each call in the stream? I thought this would be more efficient, but that was more on gut.

vinopm: That’s right, seems like no harm in creating a new distribution object each time though. We would be reusing the same random generator. From an interface point of view it seems to make more sense to return a sampled value here to match the return values in all the other methods of this class.

sam-brett:
@vinopm Thanks. Just pushed that.

vinopm:
Thanks @sam-brett . There's also https://github.com/ScottishCovidResponse/Contact-Tracing-Model/pull/39#discussion_r456376109 left I think

sam-brett:
@vinopm sorry, I'm sure GitHub occasionally hides conversations.

vinopm:
@sam-brett no worries, all looks good to me now. Thanks.

vinopm (approved):

PR #40, opened 2020-07-14
Additional stats

sam-brett:
Hi @vinopm, 

There are a couple of bits I know I need to change in this, but I'm going to be in the office tomorrow so it might be worth having a once over if you get time. 

The outputs need to be configured properly (currently hardcoded) and the testing is a little weak.

Cheers
Sam

### DIFF
--- a/src/main/java/uk/co/ramp/Outbreak.java
+++ b/src/main/java/uk/co/ramp/Outbreak.java
@@ -95,8 +95,8 @@ public class Outbreak {
       throw new InfectionMapException(message, e);
     }
 
-    try (Writer statsWriter = new FileWriter("output/stats.txt");
-        Writer rValueWriter = new FileWriter("output/rValue.csv")) {
+    try (Writer statsWriter = new FileWriter(new File(outputFolder, "stats.txt"));
+        Writer rValueWriter = new FileWriter(new File(outputFolder, "rValue.csv"))) {
       statisticsWriter.output(statsWriter, rValueWriter);
     } catch (IOException e) {
       LOGGER.error("An error occurred while outputting the statistics", e);
--- a/src/main/java/uk/co/ramp/Outbreak.java
+++ b/src/main/java/uk/co/ramp/Outbreak.java
@@ -82,10 +82,10 @@ public class Outbreak {
         diseaseProperties.randomInfectionRate() / (double) properties.timeStepsPerDay();
 
     runContactData(timeLimit * properties.timeStepsPerDay(), randomInfectionRate);
-    printOutput(timeLimit);
+    printOutput();
   }
 
-  private void printOutput(int timeLimit) {
+  private void printOutput() {
     try (Writer writer = new FileWriter(new File(outputFolder, INFECTION_MAP))) {
       new InfectionMap(population.view(), statisticsRecorder).outputMap(writer);
       eventListWriter.output();
@@ -97,7 +97,7 @@ public class Outbreak {
 
     try (Writer statsWriter = new FileWriter("output/stats.txt");
         Writer rValueWriter = new FileWriter("output/rValue.csv")) {
-      statisticsWriter.output(timeLimit, statsWriter, rValueWriter);
+      statisticsWriter.output(statsWriter, rValueWriter);
     } catch (IOException e) {
       LOGGER.error("An error occurred while outputting the statistics", e);
     }
--- a/src/main/java/uk/co/ramp/Outbreak.java
+++ b/src/main/java/uk/co/ramp/Outbreak.java
@@ -23,6 +23,8 @@ import uk.co.ramp.io.types.DiseaseProperties;
 import uk.co.ramp.io.types.OutputFolder;
 import uk.co.ramp.io.types.StandardProperties;
 import uk.co.ramp.people.VirusStatus;
+import uk.co.ramp.statistics.StatisticsRecorder;
+import uk.co.ramp.statistics.StatisticsWriter;
 
 @Service
 public class Outbreak {
@@ -36,6 +38,7 @@ public class Outbreak {
   private final LogDailyOutput outputLog;
   private final LastContactTime lastContactTime;
   private final File outputFolder;
+  private final StatisticsRecorder statisticsRecorder;
 
   private final Population population;
   private final Map<Double, CmptRecord> records = new HashMap<>();
@@ -50,7 +53,8 @@ public class Outbreak {
       EventRunner eventRunner,
       EventListWriter eventListWriter,
       LastContactTime lastContactTime,
-      OutputFolder outputFolder) {
+      OutputFolder outputFolder,
+      StatisticsRecorder statisticsRecorder) {
 
     this.population = population;
     this.diseaseProperties = diseaseProperties;
@@ -60,6 +64,7 @@ public class Outbreak {
     this.eventListWriter = eventListWriter;
     this.lastContactTime = lastContactTime;
     this.outputFolder = outputFolder.outputFolder();
+    this.statisticsRecorder = statisticsRecorder;
   }
 
   public Map<Double, CmptRecord> propagate() {
@@ -76,12 +81,21 @@ public class Outbreak {
     runContactData(timeLimit * properties.timeStepsPerDay(), randomInfectionRate);
 
     try (Writer writer = new FileWriter(new File(outputFolder, INFECTION_MAP))) {
-      new InfectionMap(population.view()).outputMap(writer);
+      new InfectionMap(population.view(), statisticsRecorder).outputMap(writer);
       eventListWriter.output();
     } catch (IOException e) {
       String message = "An error occurred generating the infection map";
       LOGGER.error(message);
       throw new InfectionMapException(message, e);
+    }
+
+    try (Writer statsWriter = new FileWriter("output/stats.txt");
+        Writer rValueWriter = new FileWriter("output/rValue.csv")) {
+      new StatisticsWriter(timeLimit, statsWriter, rValueWriter, statisticsRecorder, properties)
+          .output();
+
+    } catch (IOException e) {
+      e.printStackTrace();
     }
   }
 
--- a/src/main/java/uk/co/ramp/Outbreak.java
+++ b/src/main/java/uk/co/ramp/Outbreak.java
@@ -39,6 +39,7 @@ public class Outbreak {
   private final LastContactTime lastContactTime;
   private final File outputFolder;
   private final StatisticsRecorder statisticsRecorder;
+  private final StatisticsWriter statisticsWriter;
 
   private final Population population;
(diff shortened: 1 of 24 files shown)


src/main/java/uk/co/ramp/statistics/StatisticsRecorderImpl.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import org.springframework.stereotype.Service;
+import uk.co.ramp.people.Case;
+import uk.co.ramp.statistics.types.ImmutableInfection;
+import uk.co.ramp.statistics.types.Infection;
+
+@Service
```
vinopm: This is not necessary as we are initialising StatisticsRecorderImpl in StatisticsRecorderContext as a bean

src/main/java/uk/co/ramp/statistics/StatisticsWriter.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import org.springframework.stereotype.Service;
+import uk.co.ramp.io.csv.CsvWriter;
+import uk.co.ramp.io.types.StandardProperties;
+import uk.co.ramp.statistics.types.ImmutableRValueOutput;
+import uk.co.ramp.statistics.types.Infection;
+
+@Service
```
vinopm: This should be removed as StatisticsWriter is already defined as a bean in StatisticsRecorderContext. I had to comment this line out to run the model without a failure

src/main/java/uk/co/ramp/statistics/StatisticsRecorder.java
```diff
@@ -0,0 +1,24 @@
...
+
+  void recordInfectionSpread(Case seed, int infections);
+
+  Map<Integer, Integer> getContactsTraced();
+
+  Map<Integer, Integer> getPeopleInfected();
+
+  Map<Integer, Integer> getPersonDaysIsolation();
+
+  Map<Integer, List<Infection>> getR0Progression();
```
vinopm, line 26: These should probably not be interface methods as they are only internally used within the statistics package. We can remove these

src/main/java/uk/co/ramp/statistics/StatisticsRecorderImpl.java
```diff
@@ -0,0 +1,66 @@
...
+    return peopleInfected;
+  }
+
+  public Map<Integer, Integer> getPersonDaysIsolation() {
+    return personDaysIsolation;
+  }
+
+  public Map<Integer, List<Infection>> getR0Progression() {
+    return r0Progression;
+  }
```
vinopm, line 48: See my other comment about removing these methods from the interface. Once that is done, these methods can be made package-private scope.

src/main/java/uk/co/ramp/statistics/StatisticsRecorder.java
```diff
@@ -0,0 +1,24 @@
...
+import java.util.Map;
+import uk.co.ramp.people.Case;
+import uk.co.ramp.statistics.types.Infection;
+
+public interface StatisticsRecorder {
+  void recordDaysInIsolation(int id, int duration);
+
+  void recordPeopleInfected(int time);
+
+  void recordContactsTraced(int time, int contactTraced);
```
vinopm: contactTraced -> numberOfContactsTraced would be more informative I think?

src/main/java/uk/co/ramp/statistics/StatisticsRecorder.java
```diff
@@ -0,0 +1,24 @@
...
+
+import java.util.List;
+import java.util.Map;
+import uk.co.ramp.people.Case;
+import uk.co.ramp.statistics.types.Infection;
+
+public interface StatisticsRecorder {
+  void recordDaysInIsolation(int id, int duration);
+
+  void recordPeopleInfected(int time);
```
vinopm: Change this to recordSinglePersonInfected(int time) or recordPeopleInfected(int time, int noOfPeople)

src/main/java/uk/co/ramp/statistics/StatisticsRecorder.java
```diff
@@ -0,0 +1,24 @@
+package uk.co.ramp.statistics;
+
+import java.util.List;
+import java.util.Map;
+import uk.co.ramp.people.Case;
+import uk.co.ramp.statistics.types.Infection;
+
+public interface StatisticsRecorder {
+  void recordDaysInIsolation(int id, int duration);
```
vinopm: is this personId? If it is, could we change id to be more descriptive

src/main/java/uk/co/ramp/statistics/StatisticsRecorderImpl.java
```diff
@@ -0,0 +1,66 @@
...
+    Infection i = ImmutableInfection.builder().seed(seed.id()).infections(infections).build();
+
+    r0Progression.compute(
+        seed.exposedTime(),
+        (k, v) ->
+            v == null
+                ? List.of(i)
+                : Stream.of(v, List.of(i))
+                    .flatMap(Collection::stream)
+                    .collect(Collectors.toList()));
```
vinopm: Is this the same as: 
`r0Progression.computeIfAbsent(seed.exposedTime(), k -> new ArrayList<>()).add(i);` ?

src/main/java/uk/co/ramp/statistics/MovingAverage.java
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  public double getAverage() {
+    if (previousDays.isEmpty()) return 0d;
+    if (sum < 1e-6) return 0d;
+    double divisor = previousDays.size();
+    return sum / divisor;
+  }
+
+  public List<Double> readData() {
```
vinopm: package-private scope + could mark with `@VisibleForTesting` annotation

src/main/java/uk/co/ramp/Outbreak.java
```diff
@@ -64,28 +66,41 @@ public Outbreak(
...
+      new InfectionMap(population.view(), statisticsRecorder).outputMap(writer);
       eventListWriter.output();
     } catch (IOException e) {
       String message = "An error occurred generating the infection map";
       LOGGER.error(message);
       throw new InfectionMapException(message, e);
     }
+
+    try (Writer statsWriter = new FileWriter("output/stats.txt");
+        Writer rValueWriter = new FileWriter("output/rValue.csv")) {
```
vinopm: locations to be picked up from config

src/main/java/uk/co/ramp/statistics/StatisticsWriter.java
```diff
@@ -0,0 +1,70 @@
...
+        rValueOutputs.add(
+            ImmutableRValueOutput.builder()
+                .time(i)
+                .newInfectors(orDefault.size())
+                .newInfections(seeded)
+                .r(r)
+                .sevenDayAverageR(movingAverage.getAverage())
+                .build());
+      }
+    }
```
vinopm: I feel like this logic would sit better within the statisticsRecorder. We could have a new method for the rolling avg

vinopm (changes requested):

src/main/java/uk/co/ramp/statistics/StatisticsRecorderImpl.java
sam-brett: If we do that, then the StatisticsWriter class needs the Impl rather than the interface, which would also affect the Context... unless I'm missing something... its been a long day so thats possible.

src/main/java/uk/co/ramp/statistics/StatisticsRecorderImpl.java
sam-brett: Haha... yep.. thats simpler.

src/main/java/uk/co/ramp/statistics/MovingAverage.java
sam-brett: I did not know that annotation existed... that's great.

src/main/java/uk/co/ramp/statistics/StatisticsWriter.java
sam-brett: Moved, but will need interface change to expose it in statistics writer.

sam-brett:
Thanks @vinopm. I've done most of them, but the changes to the interface impact the Context and Writer classes so I would think the get methods need to stay, but I'm running on about three brain cells so I could be wrong.

src/test/java/uk/co/ramp/event/InfectionEventProcessorTest.java
```diff
@@ -43,7 +45,9 @@
   private InfectionEventProcessor eventProcessor;
   private DiseaseProperties diseaseProperties;
 
+  @Autowired private StandardProperties properties;
   @Autowired private DistributionSampler distributionSampler;
+  @Autowired private StatisticsRecorder statisticsRecorder;
```
vinopm, line 51: Remove Autowired annotation and initialise this as: mock(StatisticsRecorder.class)

vinopm (changes requested):
Small change requested. Let's leave the issue about the interface for now, we'll clean it up later if necessary

src/main/java/uk/co/ramp/statistics/StatisticsRecorderImpl.java
```diff
@@ -0,0 +1,83 @@
...
+import uk.co.ramp.statistics.types.ImmutableInfection;
+import uk.co.ramp.statistics.types.ImmutableRValueOutput;
+import uk.co.ramp.statistics.types.Infection;
+
+public class StatisticsRecorderImpl implements StatisticsRecorder {
+
+  private final Map<Integer, Integer> personDaysIsolation = new HashMap<>();
+  private final Map<Integer, Integer> peopleInfected = new HashMap<>();
+  private final Map<Integer, Integer> contactsTraced = new HashMap<>();
+  private final Map<Integer, List<Infection>> r0Progression = new HashMap<>();
```
vinopm: I would suggest passing in the empty maps through constructor. This would improve the test readability as you would be able to pass in a populate hashmap and then call get rolling average (as an example)

vinopm (changes requested):

src/test/java/uk/co/ramp/event/InfectionEventProcessorTest.java
sam-brett: Will do this as a tech debt ticket with some of the other things we discussed yesterday like the gaussian

vinopm (approved):
Approved, but we need to make sure the time steps PR is merged before this one

PR #41, opened 2020-07-16
Test Efficacy and Tech Debt

sam-brett:
Added new input, renamed old one. Records false positive and negatives as stats. May wish to record id of person as well or who they inadvertently infect..


src/main/java/uk/co/ramp/event/AlertEventProcessor.java
```diff
@@ -93,6 +92,27 @@ public ProcessedEventResult processEvent(AlertEvent event) {
...
 
+  Optional<AlertStatus> determineTestResult(boolean isInfectious) {
+
+    double testEffectiveness = distributionSampler.uniformBetweenZeroAndOne();
+
+    if (isInfectious) {
+      if (testEffectiveness < diseaseProperties.testPositiveAccuracy()) {
+        return Optional.of(TESTED_POSITIVE);
+      } else {
+        statisticsRecorder.recordIncorrectTestResult(TESTED_NEGATIVE);
```
vinopm, line 104: Would it be worth adding stats for recordCorrectTestResult too? Otherwise we won't have a way to know the proportion of correct/incorrect test positive and negative?

src/test/java/uk/co/ramp/event/AlertEventProcessorTest.java
```diff
@@ -119,4 +132,32 @@ public void testClashingAlertStatus() {
...
+  public void determineTestResult() {
+
+    eventProcessor =
+        new AlertEventProcessor(
+            population, properties, diseaseProperties, distributionSampler, statisticsRecorder);
+
+    when(distributionSampler.uniformBetweenZeroAndOne()).thenReturn(0.99d);
+    assertThat(eventProcessor.determineTestResult(true))
+        .isNotEmpty()
+        .isEqualTo(Optional.of(TESTED_NEGATIVE));
```
vinopm: Think this reads better:
```suggestion
    assertThat(eventProcessor.determineTestResult(true)).hasValue(TESTED_NEGATIVE);
```

Could you change the other assertions below as well. Thanks!

src/test/java/uk/co/ramp/AppConfigTest.java
```diff
@@ -91,7 +91,7 @@ public void diseaseProperties() throws ConfigurationException {
     Assert.assertTrue(diseaseProperties.timeTestAdministered().max() > 0);
     Assert.assertTrue(diseaseProperties.timeTestResult().max() > 0);
 
-    Assert.assertTrue(diseaseProperties.testAccuracy() > 0);
+    Assert.assertTrue(diseaseProperties.testPositiveAccuracy() > 0);
```
vinopm, line 93: Add another assertion for testNegativeAccuracy()

src/test/java/uk/co/ramp/OutbreakTest.java
```diff
@@ -23,50 +18,40 @@
...
-
   private DiseaseProperties diseaseProperties;
 
-  public OutbreakTest() throws FileNotFoundException {}
-
   @Before
   public void setUp() throws FileNotFoundException {
     this.diseaseProperties = TestUtils.diseaseProperties();
+    StatisticsWriter statisticsWriter = mock(StatisticsWriter.class);
+    ReflectionTestUtils.setField(outbreak, "statisticsWriter", statisticsWriter);
```
vinopm: It is not necessary to mock the StatisticsWriter here. In my view this test is more of an integration/high-level style test. It uses autowired dependencies, so ends up testing the general functionality of many classes. I would suggest removing these two lines.

src/main/java/uk/co/ramp/statistics/StatisticsRecorderImpl.java
```diff
@@ -0,0 +1,97 @@
...
+    return personDaysIsolation;
+  }
+
+  public Map<Integer, List<Infection>> getR0Progression() {
+    return r0Progression;
+  }
+
+  @Override
+  public int getFalseNegatives() {
+    return incorrectTests.get(AlertStatus.TESTED_NEGATIVE);
```
vinopm: This will throw a NPE when there is no entry for TESTED_NEGATIVE in the map.
```suggestion
    return incorrectTests.getOrDefault(AlertStatus.TESTED_NEGATIVE, 0);
```

src/main/java/uk/co/ramp/statistics/StatisticsRecorderImpl.java
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  @Override
+  public int getFalsePositives() {
+    return incorrectTests.get(AlertStatus.TESTED_POSITIVE);
```
vinopm: This will throw a NPE when TESTED_POSITIVE is not in the map.
```suggestion
    return incorrectTests.getOrDefault(AlertStatus.TESTED_POSITIVE, 0);
```

vinopm (changes requested):

src/test/java/uk/co/ramp/event/AlertEventProcessorTest.java
```diff
@@ -141,23 +136,15 @@ public void determineTestResult() {
             population, properties, diseaseProperties, distributionSampler, statisticsRecorder);
 
     when(distributionSampler.uniformBetweenZeroAndOne()).thenReturn(0.99d);
-    assertThat(eventProcessor.determineTestResult(true))
-        .isNotEmpty()
-        .isEqualTo(Optional.of(TESTED_NEGATIVE));
+    assertThat(eventProcessor.determineTestResult(true)).isNotEmpty().hasValue(TESTED_NEGATIVE);
```
vinopm: isNotEmpty() is not required here

src/test/java/uk/co/ramp/OutbreakTest.java
```diff
@@ -44,14 +44,13 @@
   @Autowired private InitialCaseReader initialCaseReader;
   @Autowired private Outbreak outbreak;
   @Autowired private CompletionEventListGroup eventListGroup;
+  @Autowired private StatisticsWriter statisticsWriter;
```
vinopm: I think the tests will still pass without this line needing to be added in?

src/test/java/uk/co/ramp/event/AlertEventProcessorTest.java
```diff
@@ -141,23 +136,15 @@ public void determineTestResult() {
...
 
     when(distributionSampler.uniformBetweenZeroAndOne()).thenReturn(0.90d);
-    assertThat(eventProcessor.determineTestResult(true))
-        .isNotEmpty()
-        .isEqualTo(Optional.of(TESTED_POSITIVE));
+    assertThat(eventProcessor.determineTestResult(true)).isNotEmpty().hasValue(TESTED_POSITIVE);
```
vinopm: isNotEmpty() is not required here

src/test/java/uk/co/ramp/event/AlertEventProcessorTest.java
```diff
@@ -141,23 +136,15 @@ public void determineTestResult() {
...
 
     when(distributionSampler.uniformBetweenZeroAndOne()).thenReturn(0.90d);
-    assertThat(eventProcessor.determineTestResult(false))
-        .isNotEmpty()
-        .isEqualTo(Optional.of(TESTED_NEGATIVE));
+    assertThat(eventProcessor.determineTestResult(false)).isNotEmpty().hasValue(TESTED_NEGATIVE);
```
vinopm: isNotEmpty() is not required here

src/test/java/uk/co/ramp/event/AlertEventProcessorTest.java
```diff
@@ -141,23 +136,15 @@ public void determineTestResult() {
...
 
     when(distributionSampler.uniformBetweenZeroAndOne()).thenReturn(0.99d);
-    assertThat(eventProcessor.determineTestResult(false))
-        .isNotEmpty()
-        .isEqualTo(Optional.of(TESTED_POSITIVE));
+    assertThat(eventProcessor.determineTestResult(false)).isNotEmpty().hasValue(TESTED_POSITIVE);
```
vinopm: isNotEmpty() is not required here

src/test/java/uk/co/ramp/event/InfectionEventProcessorTest.java
```diff
@@ -38,23 +24,30 @@
...
-
   private InfectionEvent event;
 
   @Before
   public void setUp() throws FileNotFoundException {
+
+    properties = mock(StandardProperties.class);
+    when(properties.timeStepsPerDay()).thenReturn(1);
+    distributionSampler = mock(DistributionSampler.class);
+    when(distributionSampler.getDistributionValue(any())).thenCallRealMethod();
```
vinopm: Ideally we should mock the resulting behaviour here as well, but as it was being auto wired before, this change is definitely one step in the right direction. We can address this in another PR if/when needed

vinopm (dismissed):
Minor suggestions, but looks good to me

src/test/java/uk/co/ramp/event/InfectionEventProcessorTest.java
sam-brett: Something like this? 
```suggestion
        when(distributionSampler.getDistributionValue(any())).thenAnswer(i -> ((int) Math.round(((Distribution) i.getArgument(0)).mean())));

```

vinopm: @sam-brett Yes, that looks good to me. I think there are two other places in this PR where the same change can be made

sam-brett:
@vinopm did you have any other suggestions for his PR?

vinopm:
@sam-brett looks good to me

vinopm (approved):

vinopm:
@sam-brett no other suggestions for now, thanks

PR #46, opened 2020-07-31
Integration with data pipeline api for parameters

vinopm:


vinopm:
@sam-brett @johnnonweiler this works AFAIK, we may have to change the config.yaml/metadata.yaml file a bit when we try the integrated run today. Will update the PR if we end up changing the configuration, but the rest is ready to be reviewed

johnnonweiler:
@sam-brett - Merging this with the latest develop was hard.  It was so tricky that I gave up trying to merge everything, and removed the new ageDependence feature.  There were _many_ manual changes required to fix merge conflicts.  Please could you check the merge commit?

In particular, please check AlertEventProcessor.java - I changed the behaviour of timeInStatusAndTestQueue() where I was confident the old behaviour was wrong. I suspect that other bits may also have been incorrect - returning 1 day where they should be returning 1 time step.  Also, it is likely that I have made mistakes in here!

I have changed one of the toml files, so the files on this branch no longer correspond to the files which were uploaded to the data pipeline.

The new infectionRates.json parameters should probably come from the data pipeline, but I have not yet made that change.

src/main/java/uk/co/ramp/policy/isolation/BoundedDistributionSerializer.java
```diff
@@ -0,0 +1,65 @@
...
+              }
+            });
+    gen.writeEndObject();
+  }
+
+  public void write(JsonGenerator gen, Entry<String, JsonNode> entry) {
+    try {
+      gen.writeObjectField(entry.getKey(), entry.getValue());
+    } catch (IOException e) {
+      e.printStackTrace();
```
sam-brett, line 62: This exception should be handled

build.gradle
```diff
@@ -30,7 +31,9 @@ dependencies {
     implementation 'com.fasterxml.jackson.dataformat:jackson-dataformat-csv:2.11.0'
     implementation "org.immutables:value-annotations:2.8.8"
     implementation "org.immutables:gson:2.8.8"
```
sam-brett, line 33: I guess this is tech debt that I can deal with, but GSON can be removed in favour of Jackson now. GSON is only used in tests now.

src/main/java/uk/co/ramp/policy/isolation/BoundedDistributionSerializer.java
sam-brett, line 62: This exception should be handled

src/main/java/uk/co/ramp/policy/isolation/BoundedDistributionSerializer.java
```diff
@@ -0,0 +1,65 @@
...
+import org.apache.commons.math3.random.RandomGenerator;
+import uk.co.ramp.distribution.BoundedDistribution;
+import uk.ramp.mapper.DataPipelineMapper;
+
+/**
+ * This class provides a temporary serialization wrapper over the distribution from std api. This
+ * can/will be removed once we migrate ALL 'distribution' related objects write to the data
+ * pipeline.
+ */
+public class BoundedDistributionSerializer extends JsonSerializer<BoundedDistribution> {
```
sam-brett, line 22: This class isn't tested

sam-brett (approved):
@johnnonweiler Happy to approve and I'll fix the age dependency and other bits in another PR.

PR #47, opened 2020-08-17
Isolation release

sam-brett:


johnnonweiler:
I'd appreciate a PR description which explains the changes included in this PR.

It includes significant documentation changes, which I have not yet reviewed much.  There appear to be multiple versions of the documentation - for example, Build Documentation.docx and Build_Documentation.md files appear to be two versions of the same document, and they are both updated by this PR, but not to have the same contents.  Perhaps the "docx" version can be removed?

Also, the ".pptx" file is updated, and images which come from that file (I think) are updated, but the latest images have _not_ been generated from the latest ".pptx" file.

I suggest using git for version control, changing the folder structure to something other than "doc/doc_v1.0", keeping only one version of "Contact_tracing_model_overview.md", and probably also removing the document version history within documents (e.g. at the end of Build_documentation.md).

It looks like the change in SingleCaseIsolationPolicy.java is fairly independent from most of the other changes?  I will take another look at this tomorrow.

johnnonweiler:
The change in SingleCaseIsolationPolicy.java looks good.  I've pulled that change out, and suggested a very minor refactor, in #50.

PR #49, opened 2020-08-17
Adding test Capacity and App uptake features

sam-brett:


src/main/java/uk/co/ramp/event/AlertEventProcessor.java
```diff
@@ -57,7 +61,7 @@ public ProcessedEventResult processEvent(AlertEvent event) {
     }
 
     if (nextStatus.get() != event.nextStatus()) {
-      int deltaTime = timeInStatus(nextStatus.get());
+      int deltaTime = timeInStatus(nextStatus.get(), event);
```
johnnonweiler: The `timeInStatus` method is now doing more than calculating the time, (including e.g. recording statistics,) and should perhaps be renamed.

src/main/java/uk/co/ramp/event/AlertEventProcessor.java
```diff
@@ -83,6 +87,7 @@ public ProcessedEventResult processEvent(AlertEvent event) {
 
     switch (event.nextStatus()) {
       case AWAITING_RESULT:
+        //        statisticsRecorder.recordTestConducted(event.time());
```
johnnonweiler: Should this line be deleted?

src/main/java/uk/co/ramp/event/AlertEventProcessor.java
```diff
@@ -144,4 +162,15 @@ int timeInStatus(AlertStatus newStatus) {
...
+
+  private int findNextTest(int timeNow, int capacity) {
+
+    for (int testTime = timeNow; testTime < properties.timeLimit(); testTime++) {
+      if (statisticsRecorder.getTestsConducted(testTime) < capacity) {
+        statisticsRecorder.recordTestConducted(testTime);
+        return testTime - timeNow;
+      }
+    }
+    return properties.timeLimit() * properties.timeStepsPerDay() - timeNow;
```
johnnonweiler: In the loop above, `testTime` is set to `timeNow` and then compared to `properties.timeLimit()`, which suggests that `timeNow` and `properties.timeLimit()` have the same units.  This return statement suggests that they have different units (with one in days and the other in time steps).  Something must be wrong!

I think this is a good example of where it would be helpful to rename things to include their units ("Days" or "TimeSteps") in their names.  E.g. `properties.timeLimitDays()` or `nowTimeSteps`.  Alternatively, you could use a Time class everywhere, instead of integers, with `getDays()` and `getTimeSteps()`.

src/test/resources/testSeedOverride/runSettings.json
```diff
@@ -2,6 +2,7 @@
   "populationSize": 10000,
   "timeLimit": 200,
   "initialExposures": 1000,
+  "seed": 3,
```
johnnonweiler, line 5: Why was this added?  (It seems strange to specify "seed" twice in the same file - it is set on line 5 and line 7 - I would have thought this should be an error.)

src/test/java/uk/co/ramp/people/PopulationGeneratorTest.java
```diff
@@ -85,11 +84,18 @@ public void generate() {
         population.values().stream().map(Case::gender).filter(a -> a.equals(Gender.FEMALE)).count()
             / (double) popSize;
 
+    double hasApp =
+        population.values().stream().map(Case::hasApp).filter(a -> a.equals(true)).count()
```
johnnonweiler: What does `filter(a -> a.equals(true))` do?  (I expect it's useful, but it's not obvious to me!)

src/test/java/uk/co/ramp/policy/alert/AlertContactTracerTest.java
```diff
@@ -53,12 +53,29 @@ public void testPrimaryLevelTrace() {
...
+    when(population.hasApp(anyInt())).thenReturn(true);
+
     var contactTracer =
         new AlertContactTracer(tracingPolicy, eventList, population, statisticsRecorder);
 
     assertThat(contactTracer.traceRecentContacts(0, 2, 1)).containsExactlyInAnyOrder(2, 3);
   }
 
+  @Test
+  public void testPrimaryLevelTraceNoApp() {
```
johnnonweiler, line 61: This includes a lot of copy-and-pasted code from the previous test.  Could these tests be refactored to share common code?

input/infectionRates.json
```diff
@@ -0,0 +1,6 @@
+{
```
johnnonweiler: This PR builds upon the changes in PR#48 - It would be easier review if _EITHER_ it did not include those changes, _OR_ the PR targetted that branch instead of the develop branch (and then I believe that if that branch is merged into develop and deleted, GitHub will automatically switch this PR to target develop).

src/main/java/uk/co/ramp/event/AlertEventProcessor.java
```diff
@@ -115,12 +120,25 @@ public ProcessedEventResult processEvent(AlertEvent event) {
...
       case TESTED_NEGATIVE:
+        int conducted = statisticsRecorder.getTestsConducted(time);
+        int capacity =
+            (int) Math.ceil(properties.populationSize() * populationProperties.testCapacity());
+        boolean testsAvailable = conducted < capacity;
+        if (testsAvailable) {
+          statisticsRecorder.recordTestConducted(time);
+          progressionData = ImmutableMeanMax.builder().mean(1).max(1).build();
+        } else {
+          int testTime = findNextTest(time, capacity);
```
johnnonweiler, line 136: I wonder if having a queue waiting until a test is available is the right way to handle running out of test capacity - I can imagine a scenario where the backlog builds up, and eventually people are waiting more than 14 days for a test.

src/main/java/uk/co/ramp/event/AlertEventProcessor.java
```diff
@@ -115,12 +120,25 @@ public ProcessedEventResult processEvent(AlertEvent event) {
...
+          statisticsRecorder.recordTestDelayed(time, event.id());
```
johnnonweiler, line 137: The statistics recorder appears to be recording things (such as event ids) which are never used.  Why is this?

src/main/java/uk/co/ramp/io/ContactReader.java
```diff
@@ -45,6 +45,7 @@ public ContactReader(
 
   List<ImmutableContactEvent> resampleContacts(List<ImmutableContactEvent> contactEvents) {
     int steps = properties.timeStepsPerDay();
+    int offset = properties.dayOffset().orElse(0);
```
johnnonweiler, line 48: I don't know if you're generally using GitHub's "squash commits and merge" option, but it is probably a good idea here.  It would be good to mention the addition of the "dayOffset" parameter in the new commit message, as well as explaining the capacity and app uptake changes.

Also, I don't know why `input/contacts.csv` was changed in this PR, but that should be explained too.

johnnonweiler (dismissed):
This PR includes a number of changes, some which are unrelated to each other, and that makes it a bit difficult to review.  I think it's all ok.

src/main/java/uk/co/ramp/event/AlertEventProcessor.java
sam-brett: Good spot. I'll change timeLimit to timeLimitDays

src/test/resources/testSeedOverride/runSettings.json
sam-brett: I think this might have been a merge problem.

src/test/java/uk/co/ramp/people/PopulationGeneratorTest.java
sam-brett: replaced with ```population.values().stream().filter(Case::hasApp).count()```

src/test/java/uk/co/ramp/policy/alert/AlertContactTracerTest.java
sam-brett: refactored

src/main/java/uk/co/ramp/event/AlertEventProcessor.java
sam-brett: This is just a first take, Sibylle and Louise seem happy with the simple delay, but we can discuss improvements further.

src/main/java/uk/co/ramp/event/AlertEventProcessor.java
sam-brett: this is a hook in case they want to output information beyond the total number of delays.

src/main/java/uk/co/ramp/io/ContactReader.java
sam-brett: contact.csv didn't have the label field, so I added it.

src/test/resources/testSeedOverride/runSettings.json
johnnonweiler: I fear you've had another merge problem.

johnnonweiler (approved):

PR #51, opened 2020-08-19
Fixing bug where infection map was not scaled by time step.

sam-brett:


EdTownsend1 (commented):
All looks good to me.

I was trying to decide whether passing the whole StandardProperties instance is required, but I don't think it makes much difference to just passing the number of timesteps per day.

johnnonweiler (approved):
Looks good to me.

PR #53, opened 2020-08-20
Age dependent health modifier

sam-brett:
Allows the maximum health of an individual to me modified by age range.


src/main/java/uk/co/ramp/io/readers/PopulationPropertiesReader.java
```diff
@@ -34,6 +34,11 @@ public void create(Writer writer) {
             .putPopulationAges(2, ImmutableMinMax.of(25, 54))
             .putPopulationAges(3, ImmutableMinMax.of(55, 64))
             .putPopulationAges(4, ImmutableMinMax.of(65, 90))
+            .putAgeDependence(0, 0.9)
+            .putAgeDependence(1, 0.6)
+            .putAgeDependence(2, 0.2)
+            .putAgeDependence(3, 0.8)
+            .putAgeDependence(4, 0.4)
```
johnnonweiler, line 41: I don't completely understand what these "create" methods are for, but these values are different from the values in `input/populationSettings.json`.

src/main/java/uk/co/ramp/people/PopulationGenerator.java
```diff
@@ -83,4 +86,16 @@ public void setAgeRetriever(AgeRetriever ageRetriever) {
 
     return population;
   }
+
+  public double getHealthModifier(int age) {
+
+    int index =
+        properties.populationAges().entrySet().stream()
+            .filter(entry -> entry.getValue().min() <= age && entry.getValue().max() > age)
```
johnnonweiler: The age ranges are inclusive, so I believe the max should be `>=` age.  (And it would be a good idea to modify the test to confirm if I'm right.)

src/test/java/uk/co/ramp/people/PopulationGeneratorTest.java
```diff
@@ -83,19 +91,62 @@ public void generate() {
...
 
+  @Test
+  public void getHealthModifier() throws FileNotFoundException {
+    PopulationProperties populationProperties = TestUtils.populationProperties();
+
+    double max =
+        populationProperties.populationAges().values().stream()
+            .mapToDouble(MinMax::max)
+            .max()
+            .orElse(100);
```
johnnonweiler: Since we know what the populationProperties values are, I would prefer to have a simpler test, with `double max = 100;` here.  (I'd also prefer to call it `maxAge`.)

johnnonweiler (changes requested):
This looks good apart from the `>` or `>=` issue.

src/main/java/uk/co/ramp/io/readers/PopulationPropertiesReader.java
sam-brett: The create methods are just so the input files can be created from within the program. I don't think they're going to get used, but it's a pattern we started using quite early on. The numbers don't really matter as it is more so the formatting is created.

src/main/java/uk/co/ramp/people/PopulationGenerator.java
sam-brett: good spot.

johnnonweiler (approved):
Looks good!

PR #54, opened 2020-08-26
Re-adding the age dependent health

sam-brett:


sam-brett:
@johnnonweiler 

I haven't removed the GSON dependency because it's more entwined than I thought, plus it doesn't change anything for the user. I still have some work to do for the BoundedDistributionSerializer though

sam-brett:
Wow... I regret offering to write the tests for that, it was a rabbit hole!

src/main/java/uk/co/ramp/AppConfig.java
```diff
@@ -94,6 +94,20 @@ public OutputFolder outputFolder() {
...
+      return new AgeDependentHealthReader().read(reader);
+    } catch (IOException e) {
+      String message =
+          "An error occurred while parsing the age dependent health at "
+              + inputFiles().ageDependentHealth();
+      LOGGER.error(message);
+      throw new ConfigurationException(message, e);
+    }
+  }
+
```
johnnonweiler, line 110: It looks like this could be refactored to share common code.  With something along the lines of:

```
    String inputFile = inputFiles().ageDependentHealth();
    Class cls = AgeDependentHealthReader.class;
    String description = "age dependent health";
...

my_func(String inputFile, Class cls, String description) {
    try (Reader reader = getReader(inputFile)) {
      return new cls.newInstance().read(reader);
    } catch (IOException e) {
      String message =
          "An error occurred while parsing the " + description + " at " + inputFile;
      LOGGER.error(message);
      throw new ConfigurationException(message, e);
    }
```

but then again, this change is probably more effort than it's worth!

src/main/java/uk/co/ramp/io/readers/AgeDependentHealthReader.java
```diff
@@ -0,0 +1,56 @@
...
+            .addAgeDependentList(
+                ImmutableAgeDependentHealth.builder()
+                    .range(ImmutableMinMax.of(80, 100))
+                    .modifier(0.4)
+                    .build())
+            .build();
+
+    Gson gson = new GsonBuilder().setPrettyPrinting().create();
+    gson.toJson(ageDependentHealthList, writer);
+  }
```
johnnonweiler: I think Vino removed some of these "create" methods, and, as they don't seem to be used, it's probably better not to simplify the codebase and take them out.

src/main/java/uk/co/ramp/people/PopulationGenerator.java
```diff
@@ -18,6 +20,19 @@
   private PopulationProperties properties;
   private RandomDataGenerator dataGenerator;
   private AgeRetriever ageRetriever;
+  private Map<MinMax, Double> ageMap;
+
+  public static Map<VirusStatus, Integer> getCmptCounts(Map<Integer, Case> population) {
```
johnnonweiler: I don't like the name "getCmptCounts" - I'm not very clever, so I found myself thinking, "Is 'Cmpt' short for 'Component' or 'Compartment'".

src/main/java/uk/co/ramp/people/PopulationGenerator.java
```diff
@@ -18,6 +20,19 @@
...
+
+    Map<VirusStatus, Integer> pop =
+        population.values().stream()
+            .map(Case::virusStatus)
+            .collect(Collectors.groupingBy(Function.identity(), Collectors.summingInt(e -> 1)));
```
johnnonweiler, line 30: This is a style of Java that I still haven't got used to yet. :-)

Could you using `Collectors.counting()` instead of `Collectors.summingInt(e -> 1)` here?

src/main/java/uk/co/ramp/people/PopulationGenerator.java
```diff
@@ -88,15 +101,11 @@ public void setAgeRetriever(AgeRetriever ageRetriever) {
...
-    //            .findAny()
-    //            .orElse(-1);
-    //
-    //    return properties.ageDependence().getOrDefault(index, 1d);
+
+    return ageMap.entrySet().stream()
+        .filter(entry -> entry.getKey().min() <= age && entry.getKey().max() >= age)
+        .map(Map.Entry::getValue)
+        .findAny()
+        .orElse(1d);
```
johnnonweiler, line 109: This is a nice improvement on the previous version.

src/main/java/uk/co/ramp/policy/isolation/BoundedDistributionSerializer.java
```diff
@@ -38,19 +41,16 @@ public void serialize(
...
-                }
-
-                write(gen, stringJsonNodeEntry);
+            stringJsonNodeEntry -> {
+              if (stringJsonNodeEntry.getKey().equals("distribution")) {
+                ObjectNode innerDist = stringJsonNodeEntry.getValue().deepCopy();
+                innerDist.remove("rng");
+                JsonNode innerTree = objectMapper.valueToTree(innerDist);
+                write(gen, new SimpleImmutableEntry<>("distribution", innerTree));
+                return;
```
johnnonweiler: This change is another nice little improvement.

(I'd prefer an 'else' to the early return, but that's being too picky.)

src/test/java/uk/co/ramp/policy/isolation/BoundedDistributionSerializerTest.java
```diff
@@ -0,0 +1,100 @@
...
+            .max(Double.MAX_VALUE)
+            .build();
+
+    ObjectMapper mapper = new ObjectMapper();
+    SimpleModule simpleModule = new SimpleModule("SimpleModule");
+    simpleModule.addSerializer(MinMax.class, new MinMaxSerializer());
+    mapper.registerModule(simpleModule);
+
+    bds.serialize(dist, generator, mapper.getSerializerProvider());
+    assertThat(writer.toString()).contains("\"distributionValue\" : 2147483647");
```
johnnonweiler: To understand this test, it helps to know the value of Double.MAX_VALUE.  I might be inclined to use another value (but it clearly doesn't matter).

src/test/java/uk/co/ramp/policy/isolation/BoundedDistributionSerializerTest.java
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  @Test
+  public void write() throws IOException {
```
johnnonweiler, line 70: I think this `write` test is considerably more complicated than it should be, and I would simplify it.  I took a quick look at this, and came up with this:

```
  @Test
  public void write() throws IOException {
    // Move this bit to setup()
    JsonFactory factory = new JsonFactory();
    StringWriter writer = new StringWriter();
    JsonGenerator generator = factory.createGenerator(writer);
    generator.setCodec(new ObjectMapper());
    generator.useDefaultPrettyPrinter();

    // Then following setup
    String expected = "{ \"runSettings\": \"runSettings.json\" }";
    Map.Entry<String, JsonNode> entry =
        new AbstractMap.SimpleImmutableEntry<String, JsonNode>(
            "runSettings", new TextNode("runSettings.json"));

    generator.writeStartObject();
    bds.write(generator, entry);
    generator.writeEndObject();

    assertThat(writer.toString()).isEqualToNormalizingPunctuationAndWhitespace(expected);
  }
```

src/test/java/uk/co/ramp/policy/isolation/BoundedDistributionSerializerTest.java
johnnonweiler: Could use "isEqualToNormalizingPunctuationAndWhitespace" here, rather than contains?

johnnonweiler: (Or could you easily make the test work with  "isEqualTo"?)

johnnonweiler (dismissed):
All looks good.  You could ignore all my comments, but personally I would be inclined to simplify that last "write()" test.

src/main/java/uk/co/ramp/io/readers/AgeDependentHealthReader.java
johnnonweiler: Reading that comment back, I don't think it makes sense - I was trying to suggest that it _would_ be good to simplify the codebase.

src/main/java/uk/co/ramp/AppConfig.java
sam-brett: I see both sides, but I'll leave it as it is. They aren't very difficult to follow.

src/main/java/uk/co/ramp/io/readers/AgeDependentHealthReader.java
sam-brett: Originally we were going to have the code be able to generate it's own default inputs if it was run in an empty folder, but the pipeline has removed the need for that. I'll see how hard it is to remove them all

src/main/java/uk/co/ramp/people/PopulationGenerator.java
sam-brett: Irritatingly, counting is a long, so while it could be done, I think it would make it less clear as a simple cast wouldn't work.

src/main/java/uk/co/ramp/policy/isolation/BoundedDistributionSerializer.java
sam-brett: Changed.

src/test/java/uk/co/ramp/policy/isolation/BoundedDistributionSerializerTest.java
sam-brett: I've used isEqualToNormalizingPunctuationAndWhitespace as the printing logic can change outside of our control.

johnnonweiler (dismissed):
Thanks for the update.

(There's one very minor change that I'd like, and I'll make a quick PR to explain it.  There's no reason not to merge this one as is.)

EdTownsend1 (approved):

PR #57, opened 2020-09-08
adding timestep scaling for r-value output

sam-brett:


sam-brett:
The infection map looks to be working fine. This change is just for the r-value csv

EdTownsend1 (changes requested):
There are still two edits that need to be made in the _getRollingAverage_ method then I'm happy.

`MovingAverage movingAverage = new MovingAverage(period * properties.timeStepsPerDay());`

and

`for (int i = 0; i < properties.timeLimitDays() * properties.timeStepsPerDay(); i++)`

EdTownsend1 (approved):
Looks good to me.

